PR #6, opened 2023-08-28
SDP-829: Set pw flow instead of forgot pw

ceciliaromao:
Separated password flow from forgot password, for the case the user is logged in and wants to reset password.

### DIFF
--- a/src/pages/SetNewPassword.tsx
+++ b/src/pages/SetNewPassword.tsx
@@ -1,0 +1,193 @@
+import { useEffect, useState } from "react";
+import {
+  Heading,
+  Input,
+  Button,
+  Notification,
+  Link,
+} from "@stellar/design-system";
+// import { useDispatch } from "react-redux";
+import { useNavigate } from "react-router-dom";
+
+// import { AppDispatch } from "store";
+// import {
+//   resetForgotPasswordAction,
+//   resetPasswordAction,
+// } from "store/ducks/forgotPassword";
+import { useRedux } from "hooks/useRedux";
+
+export const SetNewPassword = () => {
+  // const dispatch: AppDispatch = useDispatch();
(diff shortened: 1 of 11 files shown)


src/pages/SetNewPassword.tsx
```diff
@@ -0,0 +1,217 @@
...
+      setCurrentPassword("");
+      setNewPassword("");
+      setConfirmNewPassword("");
+    }
+  }, [forgotPassword.status]);
+
+  const handleSignOut = () => {
+    if (USE_SSO) {
+      // reset user store (from session storage)
+      singleUserStore().then();
```
piyalbasu, line 43: what does the `.then()` do here?

src/pages/SetNewPassword.tsx
```diff
@@ -0,0 +1,217 @@
...
+    event.preventDefault();
+    handleSignOut();
+    navigate("/");
+  };
+
+  const validatePassword = () => {
+    setErrorPassword(currentPassword ? "" : "Current password is required");
+  };
+
+  const validateNewPassword = () => {
```
piyalbasu, line 61: this looks pretty similar to the password validation in `src/pages/ResetPassword.tsx`. It'd be great to move this logic into a helper to prevent duplication, if possible

src/pages/SetNewPassword.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+    setErrorNewPassword(errorMsg);
+
+    if (confirmNewPassword) {
+      validatePasswordMatch();
+    }
+  };
+
+  const validatePasswordMatch = () => {
+    if (confirmNewPassword) {
```
piyalbasu, line 85: same with this logic

piyalbasu (approved):
🚀 🚀

PR #13, opened 2023-09-25
 [SDP-781] State refactor: manage session + payment details

quietbits:
This PR is a bit larger than I would like, but I ran into some issues with RTK + TSQ on Payment Details, so I had to include a details update as well.

## Updated
- Payment details
- Retry failed payment
- Session + token refresh (for TSQ queries only)

## General
- `fetchApi` helper method that acts like a middleware for TSQ queries. We're handling sessions and errors here to avoid doing this in every call separately. Simply use `fetchApi` instead of `fetch`.
- Created the `apiQueries` directory where TanStack Query hooks should go. Because they are reusable, it's better to have them outside the components (as we had them before). The code will also scale better structured this way.
- Removing no longer needed files from `api` and `ducks` directories after converting them to TSQ hooks.
- Removed `gitInfo.ts` from the repo. This file is generated and is causing a lot of unnecessary merge conflicts, even for small changes. We don't need to have it in the repo, but it's still functioning as before with builds.
- Added TanStack Query devtools for debugging (runs only in dev).

## Session
- Created the `localStorageSessionToken` helper method for easier handling of the session token in the Local Storage.
- Also, created a custom event (`SESSION_EXPIRED_EVENT`) to handle session expired on the UI in a more framework-agnostic way. Since we're storing the session token in the Local Storage anyway, there is really no need to have it in Redux or another app state.
- Added token expiration check in the `fetchApi` helper to optimize token refresh. Update it only if it expires in less than 5 min (our tokens are valid for 15 min) instead of on every API call. We can change the 5 min threshold, but that felt like a good number from the UX perspective.

## To improve
We'll need to create a wrapper component to handle loading states after we're done refactoring the state. That way, we'll better understand how to handle it and make it consistent throughout the app.


src/helpers/fetchApi.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (minRemaining <= 0) {
+      sessionExpired();
+    } else if (minRemaining < 5) {
+      token = await refreshToken(token);
+    }
+
+    config.headers = {
+      ...config.headers,
+      Authorization: `Bearer ${token}`,
+    };
```
piyalbasu, line 41: Very NIT: should we include
```
        headers: {
          "Content-Type": "application/json",
        },
```
here? I would guess it'd be an edge case to have to request a different type of header

src/components/RetryFailedPayment.tsx
```diff
@@ -1,42 +1,34 @@
...
-  const { isFetching, data, isError, error, refetch } = useQuery({
-    queryKey: ["RetryFailedPayment"],
-    queryFn: retryPayment,
-    enabled: false,
-  });
-
-  const isSuccess = Boolean(data);
+  const { isFetching, data, isError, isSuccess, error, refetch } =
+    usePaymentsRetry([paymentId]);
+  const queryClient = useQueryClient();
```
piyalbasu, line 20: this is so much nicer! 🎉

piyalbasu (approved):

PR #19, opened 2023-10-05
SDP-897: Remove Stellar logo from Page Header

ceciliaromao:
It was decided to not use the Stellar logo in the product, since this will be hosted by others and if we were to use the Stellar logo, it could seem like we’re co-responsible for their use case, which could raise a legal problem for SDF.

### DIFF
--- a/src/generated/gitInfo.ts
+++ b/src/generated/gitInfo.ts
@@ -1,1 +1,1 @@
-export default { commitHash: "d13cf63", version: "1.0.0-rc2-10-gd13cf63" };
+export default { commitHash: "4aaf371", version: "1.0.0-rc2-21-g4aaf371" };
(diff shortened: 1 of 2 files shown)


CaioTeixeira95 (approved):
LGTM

src/generated/gitInfo.ts
```diff
@@ -1 +1 @@
-export default { commitHash: "4aaf371", version: "1.0.0-rc2-21-g4aaf371" };
+export default { commitHash: "d13cf63", version: "1.0.0-rc2-10-gd13cf63" };
```
quietbits, line 1: `src/generated/gitInfo.ts` shouldn't be committed to the repo anymore. It was added to the `.gitignore` file. Could you please check if `git` is not caching this locally for you? :pray: This file was causing too many unnecessary conflicts because it's autogenerated for every commit.

quietbits (approved):

PR #20, opened 2023-10-09
SDP-895: Include trustline asset issuer

ceciliaromao:
The screen that lists the trustlines doesn't show the issuer of an asset, just its code.
You can't tell if the asset is correct just by that.
It’s a basic concept of the Stellar network that an asset consists of the pair Code:Issuer. 

Screenshot 2023-10-09 at 15 10 04


quietbits (commented):
@cbshin-stellar, should we top align the sub-menu?

image

cbshin-stellar:
yep, let's top-align it @quietbits

quietbits (approved):

PR #22, opened 2023-10-10
SDP-782: State refactor: payments

quietbits:
- Replace `payments` Redux with TanStack Query
- Optimize `Pagination` component


src/components/Pagination/index.tsx
```diff
@@ -1,26 +1,27 @@
...
   maxPages,
-  onChange,
-  onBlur,
-  onPrevious,
-  onNext,
+  onSetPage,
   isLoading,
 }: PaginationProps) => {
-  const isError = currentPage > maxPages;
+  const [page, setPage] = useState<number | undefined>();
```
quietbits, line 18: Using this local `page` state variable to keep manual page input change local and trigger parent component update only on blur.

PR #36, opened 2023-10-25
[SDP-788] State refactor: users

quietbits:


src/components/NewUserModal.tsx
```diff
@@ -33,25 +44,22 @@ export const NewUserModal: React.FC<NewUserModalProps> = ({
...
+  const prevVisible = usePrevious(visible);
 
   useEffect(() => {
-    if (
-      users.newUser.status === "ERROR" ||
-      users.newUser.status === "SUCCESS"
-    ) {
-      handleClose();
+    // Clear the form when modal closes
+    if (prevVisible === true && visible === false) {
```
piyalbasu, line 54: should these vars be `isPrevVisible` and `isVisible` since they're booleans?

piyalbasu (approved):

src/components/NewUserModal.tsx
piyalbasu: Also might be more succinct to do `if (isPrevVisible && !isVisible)`

quietbits: `visible` is a prop we use in the `Modal` component to stay consistent with how they are named in native HTML elements (for example, `disabled` in button). We could name `isPrevVisible`, but I tried to keep it consistent. Is it confusing?

quietbits: I kept the longer form to make it clear that we're checking booleans. :woman_shrugging: Not sure if staying consistent with props is better in this case. :thinking:

piyalbasu: Imo, I think we should use the common boolean naming convention of `is*/has*`. I don't think we should make our code more confusing bc HTML uses bad naming conventions 😜

quietbits: I'll update to `isVisible` in this case, but we shouldn't make changes in the SDS because the idea is that we're extending the native HTML elements. It would require many manual updates to include `isDisabled`, `isChecked`, etc. And I'm sure we would miss some of them. :sweat_smile:

PR #53, opened 2023-11-29
[SDP-960] Add dropdown for choosing Verification Type when creating new disbursements

ceciliaromao:
Screenshot 2023-11-29 at 12 25 45


src/components/DisbursementDetails/index.tsx
```diff
@@ -81,10 +85,17 @@ export const DisbursementDetails: React.FC<DisbursementDetailsProps> = ({
     isFetching: isWalletAssetsFetching,
   } = useAssetsByWallet(details.wallet.id);
 
+  const {
+    data: verificationsTypes,
+    error: verificationsTypesError,
+    isFetching: isVerificationTypesFetching,
```
quietbits, line 91: Nit: naming is a bit inconsistent. We should either use `verificationTypes...` (singular `verification`) or `verificationsTypes` (plural `verifications`).

src/helpers/formatVerificationFieldString.ts
```diff
@@ -0,0 +1,14 @@
+export const formatVerificationFieldString = (input: string) => {
```
quietbits, line 1: It looks like `input` can be an empty string/optional (`{formatVerificationFieldString(details.verificationField ?? "")}`). Should we reflect that here? That way, we could optimize by returning the same input early if it is an empty string.

quietbits: image

The programmatic formatting is not great as we're not making uppercase words that should be (for example, PIN and ID). Maybe we should do manual mapping to labels instead? It would also be better for translations if they are added later. :thinking: What do you think?

Something like:
```json
{
  "DATE_OF_BIRTH": "Date of Birth",
  "PIN": "PIN",
  "NATIONAL_ID_NUMBER": "National ID Number"
}
```

quietbits (approved):

src/helpers/formatVerificationFieldString.ts
ceciliaromao: Great idea! Thanks @quietbits

PR #59, opened 2023-12-21
[SDP-1024] Display external payment ID in Payment Details page

ziyliu:
Displays an "External Payment ID" on the payments page if one is valid. The id can be variable length depending on the org so we will displayed a shortened string for id's that are longer than 20 characters.

id longer than 20 chars
Screenshot 2024-01-02 at 3 30 03 PM

id less than 20 chars
Screenshot 2024-01-02 at 3 30 15 PM

no id
Screenshot 2024-01-02 at 3 31 51 PM

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -339,6 +339,7 @@ export type PaymentDetails = {
   assetCode: string;
   status: string;
   statusHistory: PaymentDetailsStatusHistoryItem[];
+  externalPaymentId: string;
 };
 
 // =============================================================================
@@ -616,6 +617,7 @@ export type ApiPayment = {
   receiver_wallet: ApiPaymentReceiverWallet;
   created_at: string;
   updated_at: string;
+  external_payment_id: string;
 };
 
 export type ApiPayments = {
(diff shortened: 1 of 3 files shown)


ceciliaromao (approved):
LGTM

src/types/index.ts
```diff
@@ -338,6 +338,7 @@ export type PaymentDetails = {
   totalAmount: string;
   assetCode: string;
   statusHistory: PaymentDetailsStatusHistoryItem[];
+  externalPaymentId: string;
```
quietbits, line 341: Should we make it optional?

```suggestion
  externalPaymentId?: string;
```

src/types/index.ts
```diff
@@ -615,6 +616,7 @@ export type ApiPayment = {
   receiver_wallet: ApiPaymentReceiverWallet;
   created_at: string;
   updated_at: string;
+  external_payment_id: string;
```
quietbits, line 619: Same as above. It might be helpful to indicate that we might not always have the payment ID.

```suggestion
  external_payment_id?: string;
```

quietbits (approved):

src/types/index.ts
ziyliu: oooh yes!

PR #60, opened 2024-01-02
[SDP-1026]: Add a cancel button to Payments Details page

ceciliaromao:


src/pages/PaymentDetails.tsx
```diff
@@ -59,6 +78,38 @@ export const PaymentDetails = () => {
...
+  ) => {
+    event?.preventDefault();
+    setIsModalVisible(false);
+  };
+
+  const handleCancelPayment = (
+    event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+  ) => {
+    event.preventDefault();
+    cancelPayment({ paymentId: formattedPayment?.id });
```
quietbits, line 99: I think it's pretty safe to assume that we will always have payment ID here (otherwise the cancel button wouldn't be rendered). We can add a check here to make TypeScript happy. What do you think? :nerd_face:

src/pages/PaymentDetails.tsx
```diff
@@ -84,6 +149,16 @@ export const PaymentDetails = () => {
...
+
+                <Button
+                  variant="error"
+                  size="sm"
+                  icon={<Icon.Block />}
+                  onClick={showModal}
+                  disabled={isCanceled}
+                >
+                  Cancel
+                </Button>
```
quietbits, line 161: Can a successful payment be canceled?

quietbits (approved):

PR #62, opened 2024-01-10
[SDP-1039] Environment variables: Support process.env along with window._env_

quietbits:
- Added support for `process.env` variables.
- Moved all env variables to a separate file `src/constants/envVariables.ts` for convenience.


src/store/index.ts
```diff
@@ -19,14 +19,6 @@ import { reducer as userAccount } from "store/ducks/userAccount";
 export type RootState = ReturnType<typeof store.getState>;
 export type AppDispatch = typeof store.dispatch;
 
-const loggerMiddleware =
-  (storeVal: any) => (next: any) => (action: Action<any>) => {
-    console.log("Dispatching: ", action.type);
-    const dispatchedAction = next(action);
-    console.log("NEW STATE: ", storeVal.getState());
-    return dispatchedAction;
-  };
```
quietbits, line 28: We're not really using these logs.

ceciliaromao (approved):
LGTM

piyalbasu (approved):

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "@stellar/tsconfig",
   "compilerOptions": {
     "baseUrl": "src/",
-    "target": "es5",
+    "target": "es2017",
```
quietbits, line 5: This is the minimum target that supports the top-level `await` that we need for the `envVariables.ts` export.

src/constants/envVariables.ts
```diff
@@ -24,24 +24,52 @@ declare global {
...
-  process?.env?.REACT_APP_OIDC_REDIRECT_URI || window._env_.OIDC_REDIRECT_URI;
-export const OIDC_SCOPE =
-  process?.env?.REACT_APP_OIDC_SCOPE || window._env_.OIDC_SCOPE;
-export const OIDC_USERNAME_MAPPING = (process?.env
-  ?.REACT_APP_OIDC_USERNAME_MAPPING ||
-  window._env_.OIDC_USERNAME_MAPPING) as OidcUsername;
+const WINDOW_ENV_PATH = "/settings/env-config.js";
+
+const generateEnvConfig = async () => {
+  if (process?.env?.REACT_APP_DISABLE_WINDOW_ENV !== "true") {
```
quietbits, line 30: To avoid loading `env-config.js` if it's not used (it will get a 404 error if the file is not on the server), we have to load it dynamically.

src/constants/envVariables.ts
```diff
@@ -24,24 +24,52 @@ declare global {
...
+    const response = await fetch(WINDOW_ENV_PATH);
```
piyalbasu, line 31: could you evaluate `response` to see if the file exists? If it doesn't, you could just return early.

If you're able to do that, you might not need the additional flag of `process?.env?.REACT_APP_DISABLE_WINDOW_ENV` at all

PR #66, opened 2024-01-23
[SDP-1015]: Add SMS preview & editing before sending a new disbursement

ceciliaromao:
Adds the ability to trigger disbursements for multiple organizations. We want to have a custom SMS per organization. 

- As a user, I want to see (preview) the invitation SMS before sending a disbursement to ensure the content is correct. 
- As a user, I want the ability to edit the invitation SMS before sending a disbursement to customize the message depending on the campaign or organization.


src/components/DisbursementInviteMessage/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+}: DisbursementInviteMessageProps) => {
+  console.log(draftMessage);
+  enum radioValue {
+    ORGANIZATION = "organization",
+    CUSTOM = "custom",
+  }
+
+  const standardOrgMessage =
+    "You have a payment waiting for you from the {{.OrganizationName}}. Click {{.RegistrationLink}} to register.";
+  const customMessagePlaceholder =
```
quietbits, line 30: It looks like this placeholder is used as a value instead of a placeholder. If we want to use it as a value, we need to make sure the value will be shown correctly when switching between default and custom messages. Right now, it works only on the first switch.

src/components/DisbursementInviteMessage/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+                <form className="ReceiverInviteMessage__form">
+                  <Textarea
+                    fieldSize="sm"
+                    id="textarea-custom-input"
+                    rows={5}
+                    value={customMessageInput}
+                    onChange={(event) => {
+                      setCustomMessageInput(event.target.value);
+                      updateCustomMessageInput(event);
+                    }}
```
quietbits, line 119: We're calling `setCustomMessageInput` twice (it's included in `updateCustomMessageInput` method). If the custom message is updated only in one place, we could simplify the flow by doing everything directly here. The more functions we add, the higher the chance we'll miss something, so I'd try to avoid that if possible.

src/components/DisbursementInviteMessage/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+          {isEditMessage ? (
+            <>
+              <div className="ReceiverInviteMessage__options">
+                <RadioButton
+                  id="msg-std"
+                  name="receiver-message"
+                  label={
+                    <Fragment key="msg-std-label">
+                      {"Default message"}{" "}
+                    </Fragment>
```
quietbits, line 92: Why do we need this `Fragment` here?

src/components/DisbursementInviteMessage/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+  }, [dispatch]);
+
+  const handleOptionChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (event.target.value === radioValue.ORGANIZATION) {
+      setCustomMessageInput("");
+      setSelectedOption(radioValue.ORGANIZATION);
+    } else {
+      setSelectedOption(radioValue.CUSTOM);
+    }
+  };
```
quietbits, line 52: I don't see where we're updating the parent message (`onChange`) when switching from custom to default message. We probably should clear the custom message when that happens. What do you think?

src/components/DisbursementInviteMessage/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+            disbursement. It contains a link to the appropriate wallet at the
+            end of the message. Please check all values for accuracy.
+          </div>
+
+          {isEditMessage ? (
+            <>
+              <div className="ReceiverInviteMessage__options">
+                <RadioButton
+                  id="msg-std"
+                  name="receiver-message"
```
quietbits, line 88: Should we change the name to include disbursement? This is the same name we have for the `ReceiverInviteMessage` component options.

src/components/DisbursementInviteMessage/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+                id="textarea-standard"
+                disabled
+                rows={5}
+                value={
+                  draftMessage ??
+                  organization.data.smsRegistrationMessageTemplate ??
+                  standardOrgMessage
+                }
+              ></Textarea>
+            </div>
```
quietbits, line 150: Is this the same component as above on line 123 (`id` is the same)? If it is the same component with a different value, we should create a separate component (in this file) to render from a single place to avoid duplicating code.

src/components/DisbursementInviteMessage/styles.scss
```diff
@@ -0,0 +1,30 @@
+@use "./styles-utils.scss" as *;
```
quietbits, line 1: This file is exactly the same as `src/components/ReceiverInviteMessage/styles.scss`. We should remove it. It looks like the `DisbursementInviteMessage` component is using the same styles.

reecexlm:
can you add PR description as well.

quietbits (approved):

PR #68, opened 2024-01-26
[SDP-1042]: display `Created By` and `Started By` in disbursement details screen

ceciliaromao:


src/helpers/formatDisbursements.ts
```diff
@@ -12,6 +12,8 @@ export const formatDisbursement = (
   id: disbursement.id,
   name: disbursement.name,
   createdAt: disbursement.created_at,
+  createdBy: `${disbursement.created_by.first_name} ${disbursement.created_by.last_name}`,
+  startedBy: `${disbursement.started_by.first_name} ${disbursement.started_by.last_name}`,
```
quietbits, line 16: If `createdBy` and `startedBy` are optional, should we check if `disbursement.created_by.first_name` and `disbursement.created_by.last_name` (same for `started_by`) are set and only then assign the value? If they are not always returned from the backend, this code will break.

Something like:
```typescript
disbursement.created_by?.first_name && disbursement.created_by?.last_name ? `${disbursement.created_by?.first_name} ${disbursement.created_by?.last_name}` : null
```

src/types/index.ts
```diff
@@ -533,6 +535,16 @@ export type ApiDisbursement = {
...
+  created_by: {
+    id: string;
+    first_name: string;
+    last_name: string;
+  };
+  started_by: {
+    id: string;
+    first_name: string;
+    last_name: string;
+  };
```
quietbits, line 547: If `createdBy` and `startedBy` can be optional (lines 213 and 214 above), does it mean these could also be optional?

quietbits (commented):
The update looks good. Just added some comments to clarify types. :pray:

src/pages/DisbursementDetails.tsx
```diff
@@ -194,6 +194,20 @@ export const DisbursementDetails = () => {
                 </CopyWithIcon>
               </div>
             </div>
+
+            <div className="StatCards__card__item">
+              <label className="StatCards__card__item__label">Created by</label>
+              <div className="StatCards__card__item__value">
+                {disbursementDetails.details.createdBy}
```
quietbits, line 201: We might need to add a fallback here to make sure UI doesn't display `undefined` (same below).
```suggestion
                {disbursementDetails.details?.createdBy ?? ""}
```

quietbits (approved):

PR #76, opened 2024-02-16
[SDP-1068] Display future balance

CaioTeixeira95:
This PR adds the `Future Balance` label in the `DisbursementDetails` component to display the balance for the asset on the distribution account after the disbursement is completed.

### DIFF
--- a/src/components/DisbursementDetails/index.tsx
+++ b/src/components/DisbursementDetails/index.tsx
@@ -25,6 +25,7 @@ import "./styles.scss";
 
 interface DisbursementDetailsProps {
   variant: DisbursementStep;
+  assetBalance?: string;
   details?: Disbursement;
   csvFile?: File;
   onChange?: (state: Disbursement) => void;
@@ -50,12 +51,14 @@ const initDetails: Disbursement = {
   createdAt: "",
   status: "DRAFT",
   statusHistory: [],
-  smsRegistrationMessageTemplate: ""
+  smsRegistrationMessageTemplate: "",
+  stats: undefined,
 };
 
 export const DisbursementDetails: React.FC<DisbursementDetailsProps> = ({
   variant,
   details = initDetails,
+  assetBalance,
   csvFile,
   onChange,
   onValidate,
@@ -241,6 +244,13 @@ export const DisbursementDetails: React.FC<DisbursementDetailsProps> = ({
             <label className="Label Label--sm">Disbursement name</label>
             <div className="DisbursementDetailsFields__value">
               {details.name}
+            </div>
+          </div>
+
+          <div>
+            <label className="Label Label--sm">Future Balance</label>
+            <div className="DisbursementDetailsFields__value">
+              {Number(assetBalance) - Number(details.stats?.totalAmount)}
             </div>
           </div>
 
--- a/src/pages/DisbursementDraftDetails.tsx
+++ b/src/pages/DisbursementDraftDetails.tsx
@@ -57,6 +57,9 @@ export const DisbursementDraftDetails = () => {
   const [isDraftInProgress, setIsDraftInProgress] = useState(false);
   const [isResponseSuccess, setIsResponseSuccess] = useState<boolean>(false);
 
+  const allBalances = organization.data.assetBalances?.[0].balances;
+
+  console.log(organization.data.assetBalances);
   const dispatch: AppDispatch = useDispatch();
   const navigate = useNavigate();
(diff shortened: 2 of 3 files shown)


src/components/DisbursementDetails/index.tsx
```diff
@@ -244,6 +247,17 @@ export const DisbursementDetails: React.FC<DisbursementDetailsProps> = ({
             </div>
           </div>
 
+          <div>
+            <label className="Label Label--sm">Future balance</label>
+            <div
+              className={`DisbursementDetailsFields__value ${
+                futureBalance >= 0 ? "" : "DisbursementDetailsFields__negative"
```
quietbits, line 254: If `futureBalance` is always a `number`, we can do strict equals `>==`.

src/components/DisbursementDetails/index.tsx
```diff
@@ -244,6 +247,17 @@ export const DisbursementDetails: React.FC<DisbursementDetailsProps> = ({
...
+              }`}
+            >
+              {futureBalance}
```
quietbits, line 257: Should we format this asset amount?  The `AssetAmount` component can do that.

src/pages/DisbursementDraftDetails.tsx
```diff
@@ -176,6 +181,14 @@ export const DisbursementDraftDetails = () => {
     resetState();
   };
 
+  const handleCalculateFutureBalance = (): number => {
+    const assetBalance = Number(
+      allBalances?.find((a) => a.assetCode === draftDetails?.details.asset.code)
+        ?.balance,
+    );
+    return assetBalance - Number(draftDetails?.details.stats?.totalAmount);
```
quietbits, line 189: To make sure we don't get NaN if `totalAmount` is undefined.

```suggestion
    return assetBalance - Number(draftDetails?.details.stats?.totalAmount || 0);
```

src/pages/DisbursementDraftDetails.tsx
```diff
@@ -231,7 +244,8 @@ export const DisbursementDraftDetails = () => {
         }}
         isDraftDisabled={!isCsvFileUpdated}
         isSubmitDisabled={
-          !(Boolean(draftDetails) && Boolean(csvFile) && canUserSubmit)
+          !(Boolean(draftDetails) && Boolean(csvFile) && canUserSubmit) ||
+          handleCalculateFutureBalance() < 0
```
quietbits, line 248: We're calling the `handleCalculateFutureBalance()` function multiple times. How do you feel about assigning it to a variable and using that variable instead?

src/pages/DisbursementsNew.tsx
```diff
@@ -155,9 +157,43 @@ export const DisbursementsNew = () => {
...
+            ...draftDetails?.stats,
+            totalAmount: totalAmount?.toString() ?? "0",
+          },
+        } as Disbursement);
+
+        // update future balance
+        const assetBalance = allBalances?.find(
+          (a) => a.assetCode === draftDetails?.asset.code,
+        )?.balance;
+        setFutureBalance(Number(assetBalance) - totalAmount!);
```
quietbits, line 191: It might be safer to check that we have the `totalAmount` instead of assuming it will be there. What do you think?

```suggestion
     if (totalAmount) {
        setFutureBalance(Number(assetBalance) - totalAmount);
      }
```

quietbits (commented):
Looks good! :raised_hands:  I just have a few small comments/suggestions.

:question: Should we use `BigNumber` for `futureBalance` calculations? If we're dealing with precision greater than 2 decimal points, we could run into some issues with the "normal" calculations. Thoughts? :pray:

src/components/DisbursementDetails/index.tsx
CaioTeixeira95: I don't think so since we are showing the asset selected above. What do you think?

src/components/DisbursementDetails/index.tsx
quietbits: Ahh, my bad. No need to check for type here.

src/components/DisbursementDetails/index.tsx
quietbits: I think we should at least format the amount to match the rest of the app. We could omit the asset code. I think it's more clear which balance we're talking about there if the asset code is included, though.

Before
image

After
image

quietbits (approved):
Looks good! :rocket:

Just a nitpick about the order of imports. We try to group them like this: libraries > absolute paths > types > styles

For example:
image

PR #78, opened 2024-02-29
[SDP-899] feat: patch receivers' verification info

CaioTeixeira95:
Now users can add/update the receivers' verification info. Only **NOT** confirmed verification can be updated.

### DIFF
--- a/src/apiQueries/useReceiversReceiverId.ts
+++ b/src/apiQueries/useReceiversReceiverId.ts
@@ -3,7 +3,7 @@ import { API_URL } from "constants/envVariables";
 import { fetchApi } from "helpers/fetchApi";
 import { formatPaymentReceiver } from "helpers/formatPaymentReceiver";
 import { formatReceiver } from "helpers/formatReceiver";
-import { ApiReceiver, AppError } from "types";
+import { AppError, PaymentDetailsReceiver, ReceiverDetails } from "types";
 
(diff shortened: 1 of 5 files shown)


src/apiQueries/useReceiversReceiverId.ts
```diff
@@ -14,22 +14,19 @@ export const useReceiversReceiverId = <T>({
...
 }) => {
-  const query = useQuery<ApiReceiver, AppError>({
+  const query = useQuery<ReceiverDetails | PaymentDetailsReceiver, AppError>({
     queryKey: ["receivers", dataFormat, receiverId, { receiverWalletId }],
     queryFn: async () => {
-      return await fetchApi(`${API_URL}/receivers/${receiverId}`);
+      const response = await fetchApi(`${API_URL}/receivers/${receiverId}`);
+      return dataFormat === "receiver"
+        ? formatReceiver(response)
+        : formatPaymentReceiver(response, receiverWalletId);
```
quietbits, line 23: It might be safer to check if there is `receiverWalletId` because we need it in the `formatPaymentReceiver()`. The data format might be correct, but `receiverWalletId` might be missing. Or we can check for both to be 100% sure.

```suggestion
      return receiverWalletId
        ? formatReceiver(response)
        : formatPaymentReceiver(response, receiverWalletId);
```

quietbits (approved):

PR #80, opened 2024-03-13
[SDP-1121]: Improve frontend message when the distribution account doesn't exist

ceciliaromao:
The message is not helpful and you cannot read the whole public key, so it’s impossible to move on from here, unless you have access to the values from the DB. 
That happens in the “Distribution Account” screen.

Error message fixed to show entire public key.


quietbits (approved):

marwen-abid (commented):
LGTM!

PR #87, opened 2024-04-01
[SDP-1138]: Add seconds to payment history in Payment Details

ceciliaromao:
Payments are usually processed within one minute, which means the timestamps are not very meaningful. We should add seconds so a user can actually see the history and how the payment is processed on Stellar. ex: Mar 7, 2024, 12:57:06 PM

image

### DIFF
--- a/src/pages/PaymentDetails.tsx
+++ b/src/pages/PaymentDetails.tsx
@@ -182,7 +182,9 @@ export const PaymentDetails = () => {
                 <div className="PaymentDetails__wrapper">
                   <div className="PaymentDetails__info">
                     <label className="Label">Created at</label>
-                    <div>{formatDateTime(formattedPayment.createdAt)}</div>
+                    <div>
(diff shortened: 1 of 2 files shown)


src/pages/PaymentDetails.tsx
```diff
@@ -297,7 +300,7 @@ export const PaymentDetails = () => {
                       </Table.BodyCell>
                       <Table.BodyCell textAlign="right">
                         <span className="Table-v2__cell--secondary">
-                          {formatDateTime(h.updatedAt)}
+                          {formatDateTimeWithSeconds(h.updatedAt)}
```
CaioTeixeira95, line 303: Do we need to change the `Created At` as well?

CaioTeixeira95 (approved):
LGTM

src/pages/PaymentDetails.tsx
ceciliaromao: I really don't know, but I'll do so. If needed, we just have to change the function.

PR #88, opened 2024-04-03
[SDP-703]: Support the edition of the privacy_policy_link

ceciliaromao:
Screenshot 2024-04-03 at 15 54 41
Screenshot 2024-04-03 at 15 55 48
Screenshot 2024-04-03 at 15 56 05

### DIFF
--- a/src/api/patchOrgInfo.ts
+++ b/src/api/patchOrgInfo.ts
@@ -17,28 +17,31 @@ export const patchOrgInfo = async (
   }
 
   const formData = new FormData();
+  let data = {};
 
   Object.entries(fieldsToSubmit).forEach(([key, value]) => {
     switch (key) {
       case "name":
-        formData.append("data", `{"organization_name": "${value}"}`);
+        data = { ...data, organization_name: `${value}` };
         break;
       case "privacyPolicyLink":
-        formData.append("data", `{"privacy_policy_link": "${value}"}`);
+        data = { ...data, privacy_policy_link: `${value}` };
         break;
       case "timezone":
-        formData.append("data", `{"timezone_utc_offset": "${value}"}`);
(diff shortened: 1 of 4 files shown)


src/api/patchOrgInfo.ts
```diff
@@ -23,6 +23,9 @@ export const patchOrgInfo = async (
       case "name":
         formData.append("data", `{"organization_name": "${value}"}`);
         break;
+      case "privacyPolicyLink":
+        formData.append("data", `{"privacy_policy_link": "${value}"}`);
+        break;
```
quietbits, line 28: Could you please confirm that the form is saved as expected when both the name and privacy policy link are updated at the same time? I'm not sure if/how the `data` value will be merged.

ceciliaromao: That's the same problem when we want to update both the name and isApprovalRequired at the same time. But I'm not sure how to solve it. Any idea?

quietbits: We can try to use an object for `data` and then stringify it with all the props.

```typescript
  const formData = new FormData();
  let data = {};

  Object.entries(fieldsToSubmit).forEach(([key, value]) => {
    switch (key) {
      case "name":
        data = { ...data, organization_name: `${value}` };
        break;
      case "privacyPolicyLink":
        data = { ...data, privacy_policy_link: `${value}` };
        break;
      case "timezone":
        data = { ...data, timezone_utc_offset: `${value}` };
        break;
      case "isApprovalRequired":
        data = { ...data, is_approval_required: value };
        break;
      case "logo":
        formData.append("logo", value as Blob);
        break;
      default:
        throw Error(`Update organization does not accept ${key} field`);
    }
  });

  formData.append("data", `${JSON.stringify(data)}`);
```

quietbits (approved):

PR #89, opened 2024-04-03
chore - improve password error message

marwen-abid:
## What
Improve error message from reset-password screen to list the special characters accepted in the password.


src/pages/ResetPassword.tsx
```diff
@@ -100,9 +100,10 @@ export const ResetPassword = () => {
               <ul>
                 <li>at least 12 characters long,</li>
                 <li>
-                  a combination of uppercase letters, lowercase letters,
-                  numbers, and symbols.
+                  a combination of uppercase letters, lowercase letters and
+                  numbers
                 </li>
+                <li>at least one symbol from: <span style={{fontFamily: 'monospace'}}>! @ # $ % ^ &amp; *</span></li>
```
marwen-abid, line 106: @quietbits I'm not really sure what's the best way to format this message. If you have any suggestions please let me know. 

We just want to make sure that people know it's one of these symbols that count, and something like `-` doesn't.

quietbits, line 106: Ideally, we should create a CSS style instead of inline for this and apply it using a class.

quietbits (approved):

src/pages/ResetPassword.tsx
quietbits: I think this is fine. At least it's very clear what is required. I would just apply the style through class instead of inline (comment below). Otherwise, this seems good to me. :+1:

marwen-abid: addressed in c9d320bd9e113ad5d2d96e81718194a6f19e8944
Thank you for the review!

quietbits (approved):
:rocket:

PR #108, opened 2024-05-15
[SDP-1206] Make sure to only export the envs needed

marcelosalloum:
### What

Make sure to only export the envs needed

FYI: I confirmed that with the new configuration, the envs behavior still the same: they're loaded from .env if available, otherwise they are loaded from window._env_

### Why

📈 Kaizen


marwen-abid (approved):
LGTM!

PR #109, opened 2024-05-20
[SDP-1215] Disable the wallet toggle for in the WalletProviders screen unless the user has the owner role

marcelosalloum:
### What
Disable the wallet toggle for in the WalletProviders screen unless the user has the owner role.

### Owner-role view

Screenshot 2024-05-20 at 5 46 55 PM

### Non-owner view

Screenshot 2024-05-20 at 5 47 12 PM


marwen-abid (approved):
LGTM!

PR #114, opened 2024-06-24
Support Circle distribution account

quietbits:
<details>
<summary>Screenshots</summary>

Banner if missing configuration
image

Circle Distribution Account page
image

Circle Distribution Account edit configuration
image
</details>


src/hooks/useAllBalances.ts
```diff
@@ -0,0 +1,47 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isCircleAccount, isCircleAccountActive, isCircleBalancesSuccess]);
+
+  useEffect(() => {
+    if (isDone) {
+      setAllBalances(stellarBalances);
+    }
+    // Not including stellarBalances
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isDone]);
```
marcelosalloum, line 44: 💡 Should we enforce these two to be mutually exclusive?

src/components/DistributionAccountCircle.tsx
```diff
@@ -0,0 +1,294 @@
...
+import { NotificationWithButtons } from "components/NotificationWithButtons";
+
+import { AppDispatch } from "store";
+import { getOrgCircleInfoAction } from "store/ducks/organization";
+import { useRedux } from "hooks/useRedux";
+import { useCircleBalances } from "apiQueries/useCircleBalances";
+import { useUpdateCircleConfig } from "apiQueries/useUpdateCircleConfig";
+
+export const DistributionAccountCircle = () => {
+  const CIRCLE_API_MASKED_VALUE = "xxxx";
```
marcelosalloum, line 28: 💡 I think `*` is more commonly used to mask secrets:
```suggestion
  const CIRCLE_API_MASKED_VALUE = "**************";
```

src/components/DistributionAccountCircle.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+  const isSaveEnabled = Boolean(
+    (circleApiKey && circleApiKey !== CIRCLE_API_MASKED_VALUE) ||
+      (walletId && walletId !== distributionAccount?.circleWalletId),
+  );
+
+  const handleUpdate = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    updateCircleConfig();
+  };
```
marcelosalloum, line 94: 👋  As I mentioned in [this thread](https://stellarfoundation.slack.com/archives/C04C9MLM9UZ/p1719355360126979?thread_ts=1719234904.950069&cid=C04C9MLM9UZ), there's a small bug where the circle balances should be re-fetched after the circle config is updated.

marcelosalloum (approved):
LGTM!

I left a few suggestions below, the most important of them has the 👋 emoji.

Pls feel free to merge it after addressing them, or let me know if you think some of them can be left out.

src/types/index.ts
```diff
@@ -771,6 +775,11 @@ export type ApiOrgInfo = {
   sms_resend_interval: string;
   sms_registration_message_template?: string;
   payment_cancellation_period_days: string;
+  distribution_account?: {
+    circle_wallet_id?: string;
+    status: string;
+    type: string;
```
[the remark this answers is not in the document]
quietbits: Yes, let's do that to make it more clear. :+1:

src/components/DistributionAccountCircle.tsx
quietbits: Thanks for catching this! :raised_hands:

src/components/AccountBalances.tsx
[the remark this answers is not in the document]
quietbits: The `AccountBalances` component used to be specifically for Stellar accounts, but in this PR, I made it more generic so that we can use it anywhere where rendering balances is needed. It accepts array of balances and renders them, no other logic there.

We also use it in disbursements, which can be either type. But in the future, we can use it anywhere.

However, I will rename the type we currently have `StellarAccountBalance` to `AccountBalanceItem` because it's not just Stellar balance.

src/hooks/useAllBalances.ts
quietbits: I'm not sure I understand what you mean. Are you suggesting we separate Stellar and Circle balances? Having a hook like this that handles both internally is convenient because then we don't need to think about when to use which balances, for example, in disbursements.

src/components/AccountBalances.tsx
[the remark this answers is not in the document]
marcelosalloum: That makes sense to me. Thanks for explaining.

PR #132, opened 2024-08-08
[Fix] Remove draft payments from remaining payments count when calculating % successful payments

ziyliu:
When calculating the successful payments percentage, we need to remove draft payments count from the total payments count in the divisor.

for ex,
Screenshot 2024-08-10 at 8 17 06 AM
Screenshot 2024-08-10 at 8 17 19 AM

here, the `DRAFT` payment does not count towards the divisor, so the successful payment rate is 1/1 = 100%

Screenshot 2024-08-10 at 8 20 48 AM
Screenshot 2024-08-10 at 8 20 57 AM

here, a `READY` payment is added to the divisor, so the successful payment rate is 1/2 = 50%

### DIFF
--- a/src/components/DashboardAnalytics.tsx
+++ b/src/components/DashboardAnalytics.tsx
@@ -19,8 +19,8 @@ export const DashboardAnalytics = () => {
   } = useStatistics(userAccount.isAuthenticated);
 
   const calculateRate = () => {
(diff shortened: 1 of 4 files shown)


marcelosalloum (approved):
LGTM! 🚀

PR #135, opened 2024-08-15
Disable "Org Name" input when in single-tnt mode

ziyliu:
these occur when `SINGLE_TENANT_MODE: true`
login screen
Screenshot 2024-08-16 at 11 20 21 AM

forgot password 
Screenshot 2024-08-16 at 11 20 27 AM

reset password
Screenshot 2024-08-16 at 11 20 38 AM

### DIFF
--- a/public/settings/env-config.js
+++ b/public/settings/env-config.js
@@ -1,0 +1,7 @@
+window._env_ = {
+  API_URL: "https://localhost:8000",
+  STELLAR_EXPERT_URL: "https://stellar.expert/explorer/testnet",
+  HORIZON_URL: "https://horizon-testnet.stellar.org",
+  RECAPTCHA_SITE_KEY: "6Lego1wmAAAAAJNwh6RoOrsHuWnsciCTIL3NN-bn",
+  SINGLE_TENANT_MODE: false,
+};
--- a/public/settings/env-config.js
+++ b/public/settings/env-config.js
@@ -1,7 +1,0 @@
-window._env_ = {
-  API_URL: "https://localhost:8000",
-  STELLAR_EXPERT_URL: "https://stellar.expert/explorer/testnet",
-  HORIZON_URL: "https://horizon-testnet.stellar.org",
-  RECAPTCHA_SITE_KEY: "6Lego1wmAAAAAJNwh6RoOrsHuWnsciCTIL3NN-bn",
-  SINGLE_TENANT_MODE: false,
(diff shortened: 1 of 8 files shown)


public/settings/env-config.js
marcelosalloum, line 1: Is this file something we should be committing? Cc @quietbits

webpack.common.js
```diff
@@ -155,6 +155,9 @@ module.exports = {
         REACT_APP_RECAPTCHA_SITE_KEY: JSON.stringify(
           process.env.REACT_APP_RECAPTCHA_SITE_KEY || "",
         ),
+        REACT_APP_SINGLE_TENANT_MODE: JSON.stringify(
+          process.env.REACT_APP_SINGLE_TENANT_MODE || "",
+        ),
```
marcelosalloum, line 160: I'm confused. Wasn't this supposed to be a boolean?

marcelosalloum (commented):
I left a couple of questions, and I'd like to wait for Iveta's approval.

src/pages/ResetPassword.tsx
```diff
@@ -120,6 +124,7 @@ export const ResetPassword = () => {
             onChange={(e) => setOrganizationName(e.target.value)}
             value={organizationName}
             type="text"
+            disabled={SINGLE_TENANT_MODE}
```
marcelosalloum, line 127: Rather than `disabled`, we should probably change its visibility to hidden instead.

public/settings/env-config.js
quietbits: No, this file shouldn't be committed. It looks like `.gitignore` doesn't work for this file when it's modified.

src/pages/ResetPassword.tsx
quietbits: @marcelosalloum You meant that the field shouldn't be rendered, right? If we just change the visibility, there will be a gap in the markup. We need to render `null` instead.

```typescript
{ SINGLE_TENANT_MODE ? null : <Input ... /> }
```

webpack.common.js
ziyliu: values in `process.env` are always strings even if the env var has `true` or `false` set. Any value that's not `true` set to `process.env.REACT_APP_SINGLE_TENANT_MODE` will default to `false` when we generate the env configs in app!

quietbits (approved):

PR #161, opened 2024-10-15
[SDP-1355] Add a TokenSessionRefresher component to prevent the session from expiring so often when the page is open

marcelosalloum:
### What

Add a TokenSessionRefresher component to prevent the session from expiring so often when the page is open.

### Why

Address https://stellarorg.atlassian.net/browse/SDP-1355

### DIFF
--- a/src/components/SessionTokenRefresher.tsx
+++ b/src/components/SessionTokenRefresher.tsx
@@ -1,23 +1,23 @@
-import { refreshSessionToken } from "helpers/refreshSessionToken";
 import { useEffect } from "react";
 import { useDispatch } from "react-redux";
 import { AppDispatch } from "store";
 
+import { refreshSessionToken } from "helpers/refreshSessionToken";
+
 const TOKEN_REFRESH_INTERVAL = 60000; // 60,000ms = 1 minute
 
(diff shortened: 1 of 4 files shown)


src/components/SessionTokenRefresher.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  useEffect(() => {
+    // Function to call refreshSessionToken every minute
+    const tickerId = setInterval(() => {
+      try {
+        refreshSessionToken(dispatch);
+      } catch (error) {
+        console.error("Error refreshing session token: ", error);
+      }
+    }, TOKEN_REFRESH_INTERVAL);
```
quietbits, line 20: The interval is set to 2 minutes, but the comment says 1 minute. Just making sure the intent is to check every 2 minutes instead of 1.

quietbits (approved):

src/components/SessionTokenRefresher.tsx
marcelosalloum: Yeah, the intent is 2 minutes. I'll fix the comment, TY!

PR #175, opened 2024-11-05
Fix eslint configuration load that broke after PR #164

marcelosalloum:


package.json
```diff
@@ -48,6 +48,7 @@
     "eslint-plugin-prefer-arrow": "^1.2.3",
     "eslint-plugin-react": "^7.37.2",
     "eslint-plugin-react-hooks": "^5.0.0",
+    "eslint-plugin-react-refresh": "^0.4.14",
```
quietbits, line 51: Why do we need this plugin?

quietbits (approved):

package.json
marcelosalloum: Good question, I don't think it's needed TBH. I'm removing it

PR #176, opened 2024-11-05
[SDP-1369] FE changes for registration types

marcelosalloum:
### What

Initial FE changes for the new disbursement registration flow:

- [x] Remove everything that relies on country when parsing responses
- [x] Make disbursement.verification_type optional
- [x] In the “Wallets Provided” section, display a flag stating if the wallet is `user_managed` 
- [x] Update the disbursement details screen with the verification_type and registration_contact_type fields

#### Screenshots

Screenshot 2024-11-05 at 10 52 50 AM

Screenshot 2024-11-05 at 10 52 34 AM

Screenshot 2024-11-05 at 2 39 45 PM

### Why

Partially address https://stellarorg.atlassian.net/browse/SDP-1369

### Pending

EDIT: being addressed in PR #177 

- [ ] When creating a disbursement, start with the registration type
- [ ] Then Wallet provider. Which is disabled when the registration type ends with WALLET_ADDRESS
- [ ] Then Asset. `GET /assets` should be fetched without the wallet_id when the registration type ends with WALLET_ADDRESS.
- [ ] Then Verification type. Which is disabled when the registration type ends with WALLET_ADDRESS


src/helpers/formatWithTitleCase.ts
```diff
@@ -0,0 +1,13 @@
+export const formatWithTitleCase = (
+  input: string | undefined,
+): string | undefined => {
+  if (input === undefined) {
+    return undefined;
+  }
```
quietbits, line 6: It might be safer to return an empty string instead of `undefined` here.

```suggestion
  input: string | undefined,
): string => {
  if (!input) {
    return "";
  }
```

For example, if this is used in a string literal, it will output `undefined`:

```typescript
<div>{`Some text ${formatWithTitleCase(undefined)}`}</div> // Some text undefined
```

Otherwise, we'll need to do extra check where it is used:

```typescript
<div>{`Some text ${formatWithTitleCase(undefined) || ""}`}</div> // Some text
```

quietbits (approved):

src/helpers/formatWithTitleCase.ts
marcelosalloum: Cool, TY!

PR #179, opened 2024-11-06
Feat: Update GET /wallets with the `?user_managed={boolean}` query param

marcelosalloum:
- The Wallet Providers screen does not apply this query parameter and calls `GET /wallets` directly.
- The New Disbursement screen uses `GET /wallets?user_managed=false`


marwen-abid (approved):
LGTM!

PR #239, opened 2025-02-18
[SDP-1496] Improve UX on the fogotPassword->resetPassword flow

marcelosalloum:
### What

Improve UX on the fogotPassword->resetPassword flow:
- The reset-password page now parses a query parameter from the URL: `?token={confirmation-token}`
- Since the reset token is parsed from the URL, we don't need the token field anymore.

#### Now

Screenshot 2025-02-18 at 2 32 37 PM

#### Before

Screenshot 2025-02-18 at 2 33 32 PM


marwen-abid (approved):
LGTM

PR #253, opened 2025-03-17
[SDP-1526] Edit receiver info

JiahuiWho:


src/pages/ReceiverDetailsEdit.tsx
```diff
@@ -363,23 +382,29 @@ export const ReceiverDetailsEdit = () => {
...
+                        selected={selectedDob}
+                        onChange={handleDateChange}
+                        dateFormat="yyyy-MM-dd"
+                        placeholderText="Select a date"
+                        showMonthDropdown
+                        showYearDropdown
+                        scrollableYearDropdown
+                        yearDropdownItemNumber={100}
+                        maxDate={new Date()}
+                      />
```
marcelosalloum, line 406: Q: what's the adbantage of adding this new dependency instead of just setting the input type to `date`?

src/pages/ReceiverDetailsEdit.tsx
```diff
@@ -352,15 +387,23 @@ export const ReceiverDetailsEdit = () => {
...
+                              : alreadySubmittedText
                           }
-                          infoText={alreadyConfirmedText}
                         >
                           Date of Birth (Year & Month only)
                         </InfoTooltip>
                       }
                       fieldSize="sm"
-                      value={receiverEditFields.yearMonth}
+                      type="date"
```
marcelosalloum, line 405: For year-month, you should use an input with `type="month"`

src/pages/ReceiverDetailsEdit.tsx
```diff
@@ -186,9 +194,13 @@ export const ReceiverDetailsEdit = () => {
       resetUpdateState();
     }
 
+    const { name, value } = event.target;
+    // Format yearMonth input to YYYY-MM
+    const formattedValue = name === "yearMonth" ? value.substring(0, 7) : value;
+
```
marcelosalloum, line 200: If you update the year_month input to use `type="month"`, then you won't need this formatting step.

src/pages/ReceiverDetailsEdit.tsx
```diff
@@ -220,6 +232,9 @@ export const ReceiverDetailsEdit = () => {
 
     const alreadyConfirmedText =
       "This field was already confirmed by the user.";
+    const alreadySubmittedText = "This field was already submitted.";
```
marcelosalloum, line 235:
```suggestion
    const alreadySubmittedText = "This field was not confirmed by the user yet.";
```

src/pages/ReceiverDetailsEdit.tsx
```diff
@@ -352,15 +387,23 @@ export const ReceiverDetailsEdit = () => {
...
                           }
-                          infoText={alreadyConfirmedText}
                         >
                           Date of Birth (Year & Month only)
                         </InfoTooltip>
                       }
                       fieldSize="sm"
-                      value={receiverEditFields.yearMonth}
+                      type="date"
+                      max={new Date().toISOString().split("T")[0]}
```
marcelosalloum, line 406: With type month, you need to use the slice method to set the max:
```suggestion
                      max={new Date().toISOString().slice(0, 7)}
```

src/components/InfoTooltip/index.tsx
```diff
@@ -6,13 +6,15 @@ interface InfoTooltipProps {
...
   hideTooltip?: boolean;
+  showCheckMark?: boolean;
 }
 
 export const InfoTooltip = ({
   children,
   infoText,
   placement = "right",
   hideTooltip = false,
+  showCheckMark = false,
```
marcelosalloum, line 17: 💡  I think this component would be more sustainable if you replace these booleans with an `icon` property that defaults to `<Icon.Info />`.

@quietbits: thoughts?

marcelosalloum (commented):
Looking good! Here are some polishing suggestions:

src/pages/ReceiverDetailsEdit.tsx
JiahuiWho: I checked with Iveta that Firefox and Safari don't support month type ([browser support](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/month#browser_compatibility)). It's not ideal  but probably the best option for now.

src/pages/ReceiverDetailsEdit.tsx
JiahuiWho: Same as above reason. Let me know if you want to proceed with `type="month"`

src/pages/ReceiverDetailsEdit.tsx
JiahuiWho: Same as above reason. Firefox and Safari don't support month type.

src/components/InfoTooltip/index.tsx
JiahuiWho: Hmmm what do you think of using an `icon` prop that accepts a specific set of values, like "info" or "check" to limit variation and keeps a consistent UI?

marcelosalloum: I like this idea!

src/pages/ReceiverDetailsEdit.tsx
marcelosalloum: Gotcha, thanks for explaining it!

Given the market share of these browsers (image below) I agree with your current version. Let's keep `type="date"`.

Screenshot 2025-03-19 at 5 18 38 PM

src/components/InfoTooltip/index.tsx
```diff
@@ -1,32 +1,37 @@
...
   infoText,
   placement = "right",
   hideTooltip = false,
+  icon = "info",
 }: InfoTooltipProps) => {
   if (hideTooltip) {
     return <>{children}</>;
   }
 
+  const iconComponent = ICONS[icon] ?? ICONS.info;
```
quietbits, line 28: We're setting `icon` default value to `info`, so we should always get something here.

```suggestion
  const iconComponent = ICONS[icon];
```

src/pages/ReceiverDetailsEdit.tsx
```diff
@@ -335,15 +364,28 @@ export const ReceiverDetailsEdit = () => {
                       label={
                         <InfoTooltip
                           hideTooltip={
-                            !isVerificationFieldConfirmed("NATIONAL_ID_NUMBER")
+                            !isVerificationFieldSubmitted("NATIONAL_ID_NUMBER")
+                          }
+                          icon={
+                            isVerificationFieldConfirmed("NATIONAL_ID_NUMBER")
+                              ? "check"
+                              : "info"
```
quietbits, line 372: Nit: we use this and `isVerificationFieldConfirmed()` in multiple places with the same output. We could create local helper functions for them.

src/pages/ReceiverDetailsEdit.tsx
quietbits: Yeah, it would be nice if we could use `type="month"`. We could create a custom component for this consisting of two input fields. Month would be a dropdown, and year could be a number input with set range.

quietbits (approved):

src/pages/ReceiverDetailsEdit.tsx
JiahuiWho: `isVerificationFieldConfirmed()` is used in different places with different inputs (e.g., `PIN`, `NATIONAL_ID_NUMBER`), with each being duplicated twice. 

I do think the inline check keeps the code more readable. Extracting a helper function would require additional parameters to handle different cases.

PR #260, opened 2025-04-28
SDP-1610: Update receivers page for contract accounts

philipliu:
### What

When a receiver's contract account wallet is selected:

- Balances of assets registered with SDP will have their balances fetched. This feature requires RPC integration; if the `RPC_URL` configuration is not set, then the balance will show "Unavailable".
- Recent wallet history is disabled

#### `RPC_URL` is set, balances are rendered

Screenshot 2025-04-29 at 12 35 09 PM

#### `RPC_URL` is not set, balances are unavailable

Screenshot 2025-04-29 at 12 35 28 PM

This change also filters out any assets with a 0 balance from the balances field for G-addresses. This is done for consistency with contract accounts. Contract accounts do not have a concept of a trustline, and therefore, we cannot distinguish a 0 balance from never receiving a balance.

### Why

The receiver page does not work when a contract account wallet is selected.

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/src/apiQueries/useAccountBalances.ts
+++ b/src/apiQueries/useAccountBalances.ts
@@ -62,9 +62,10 @@ export const useAccountBalances = (stellarAddress: string | undefined) => {
     queryKey: ["account-balances", stellarAddress],
     queryFn: async () => {
       if (!stellarAddress) {
-        throw new Error("stellarAddress is required");
+        return [];
       }
 
+      // If the address is a Stellar address (starts with "G"), fetch from Horizon
       if (stellarAddress.startsWith("G")) {
         const accountInfo = await fetchStellarApi(
           `${HORIZON_URL}/accounts/${stellarAddress}`,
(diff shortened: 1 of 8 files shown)


src/apiQueries/useAccountBalances.ts
```diff
@@ -0,0 +1,109 @@
...
+        const rpcServer = new rpc.Server(RPC_URL);
+        const passphrase = (await rpcServer.getNetwork()).passphrase;
+        let network: Networks;
+        if (passphrase === Networks.PUBLIC) {
+          network = Networks.PUBLIC;
+        } else if (passphrase === Networks.TESTNET) {
+          network = Networks.TESTNET;
+        } else {
+          network = Networks.TESTNET;
+        }
```
marcelosalloum, line 92: Can we add Futurenet too? FYI: the backend supports futurenet

marcelosalloum (approved):
The PR LGTM but I have a design question...

Since the backend will have an RPC available at all times, do we plan to allow the FE to access that RPC? We can add a `/stellar-rpc` endpoint to our backend and forward all incoming requests to the backend-configured RPC.

I'm starting a thread on Slack to discuss that.

philipliu:
> Since the backend will have an RPC available at all times, do we plan to allow the FE to access that RPC? We can add a /stellar-rpc endpoint to our backend and forward all incoming requests to the backend-configured RPC.

That's a good point. They should be pointing to the same RPC instance. I like the idea of adding an RPC proxy endpoint since it would allow us to add SDP API authentication on top of it to prevent abuse. 

I'll leave this PR intact and address this when I start implementing the RPC integration on the backend.

PR #301, opened 2025-08-07
SDP 1796 direct payment confirmation screen and asset balance display

dpohr:
### What

-Added balance display.
-Added confirmation modal to ensure that user will not send funds accidentally.
-Changed receiver search behave, now the search triggered if 3+ chars entered.
-Added external ID tooltip.
-Changed misleading placeholders and descriptions.

image

image

image

### Why

In scope of the UX/UI testing some enhancements were requested

### Known limitations

[TODO or N/A]

### Checklist

- [ ] Ready for production


src/components/DirectPaymentCreateModal/DirectPaymentCreateModal.tsx
```diff
@@ -286,142 +309,265 @@ export const DirectPaymentCreateModal: React.FC<DirectPaymentCreateModalProps> =
...
+            <Button
+              size="sm"
+              variant="tertiary"
+              type="submit"
+              disabled={isLoading}
+              isLoading={isLoading}
+            >
+              Submit
+            </Button>
+          </Modal.Footer>
```
quietbits, line 568: I don't think these are the correct button colors. We usually have an action button purple (primary in new SDS) and a cancel button white (tertiary in new SDS). And size is larger ("md") as it was before.

image

Example from API key

image

src/components/DirectPaymentCreateModal/DirectPaymentCreateModal.tsx
```diff
@@ -286,142 +309,265 @@ export const DirectPaymentCreateModal: React.FC<DirectPaymentCreateModalProps> =
...
+
+    // Helper function to shorten wallet addresses
+    const shortenAddress = (
+      address: string,
+      startChars: number = 6,
+      endChars: number = 8,
+    ): string => {
+      if (!address || address.length <= startChars + endChars + 3) return address;
+      return `${address.slice(0, startChars)}...${address.slice(-endChars)}`;
+    };
```
quietbits, line 323: We already have a helper `shortenAccountKey.ts`. We should update it if we want it to be more flexible and use it here.

src/components/DirectPaymentCreateModal/DirectPaymentCreateModal.tsx
```diff
@@ -286,142 +309,265 @@ export const DirectPaymentCreateModal: React.FC<DirectPaymentCreateModalProps> =
...
+        ? formData.selectedReceiver.wallets.find((w) => w.wallet.id === formData.walletId)
+        : null;
+    const walletAddress = selectedWallet?.stellar_address || paymentDataToConfirm.wallet?.address;
+
+    return (
+      <div className="DirectPaymentCreateModal__confirmation">
+        <div className="DirectPaymentCreateModal__confirmationContent">
+          <div className="DirectPaymentCreateModal__confirmationWarning">
+            <p>Please review the payment details below. This action cannot be undone.</p>
+          </div>
```
quietbits, line 344: We can use `Alert` component from SDS here ([docs](https://design-system.stellar.org/alert)). The message is using color CSS variable that no longer exist. The old SDS had prefix `--color`, the new has `--sds-clr-`.

image

image

src/components/DirectPaymentCreateModal/DirectPaymentCreateModal.tsx
```diff
@@ -286,142 +309,265 @@ export const DirectPaymentCreateModal: React.FC<DirectPaymentCreateModalProps> =
...
-            <Select
-              fieldSize="sm"
-              id="assetId"
-              label="Asset"
-              value={formData.assetId}
-              onChange={handleInputChange}
-              error={formErrors.assetId}
-              required
+      <Modal.Heading>
+        {showConfirmation ? "Confirm Direct Payment" : "Create a direct payment"}
```
quietbits, line 394: Should we match the title case in both cases to make them consistent?

```suggestion
        {showConfirmation ? "Confirm Direct Payment" : "Create A Direct Payment"}
```

src/components/DirectPaymentCreateModal/DirectPaymentCreateModal.tsx
```diff
@@ -286,142 +309,265 @@ export const DirectPaymentCreateModal: React.FC<DirectPaymentCreateModalProps> =
...
-                />
-
-                {formData.selectedReceiver && !isReceiverWalletAddress && (
-                  <SelectedReceiverInfo receiver={formData.selectedReceiver} />
-                )}
-
-                {renderSearchResults()}
-              </div>
+              Confirm Payment
+            </Button>
```
quietbits, line 423: Buttons should match other modals here as well.

src/components/DirectPaymentCreateModal/styles.scss
```diff
@@ -111,6 +111,10 @@
       overflow: hidden;
       min-height: 0;
     }
+
+    &:last-child[data-visible="true"] {
+      min-height: pxToRem(85px);
+    }
```
quietbits, line 117: Curious, why do we need this?

src/components/DirectPaymentCreateModal/styles.scss
```diff
@@ -147,4 +151,75 @@
...
+
+  &__confirmationLabel {
+    font-weight: 500;
+    color: var(--color-gray-70);
+    font-size: pxToRem(14px);
+  }
+
+  &__confirmationValue {
+    font-weight: 600;
+    color: var(--color-gray-80);
```
quietbits, line 208: CSS variables starting with `--color-` are no longer valid in new SDS. All color variables now start with `--sds-clr-`. Best way to view what's available is by selecting `#root` div in dev tools. We might not have a 1:1 match from old to new, so pick the closest color (or check what is used in similar cases).

image

src/components/DirectPaymentConfirmation/DirectPaymentConfirmation.tsx
```diff
@@ -0,0 +1,88 @@
+import { Notification } from "@stellar/design-system";
+
+import { CreateDirectPaymentRequest, ApiReceiver } from "types";
+import { shortenAccountKey } from "helpers/shortenAccountKey";
```
quietbits, line 4: Nit: types import should be last

src/components/DirectPaymentConfirmation/DirectPaymentConfirmation.tsx
```diff
@@ -0,0 +1,88 @@
...
+  <div className="DirectPaymentConfirmation__row">
+    <span className="DirectPaymentConfirmation__label">{label}</span>
+    <span className="DirectPaymentConfirmation__value">{children}</span>
+  </div>
+);
+
+export const DirectPaymentConfirmation: React.FC<DirectPaymentConfirmationProps> = ({
+  paymentData,
+  selectedReceiver,
+  receiverSearch,
```
quietbits, line 29: Something is off with the search. We used to get the response back right away, but now it takes a long time and isn't consistent. Could you please take a look? It might be something to do with how it's passed as a prop. Maybe the child component isn't updating on every value change from the parent.

src/components/DirectPaymentCreateModal/DirectPaymentCreateModal.tsx
```diff
@@ -288,140 +324,192 @@ export const DirectPaymentCreateModal: React.FC<DirectPaymentCreateModalProps> =
...
+          </Modal.Body>
+          <Modal.Footer>
+            <Button size="md" variant="tertiary" type="reset" disabled={isLoading}>
+              Cancel
+            </Button>
+            <Button
+              size="md"
+              variant="primary"
+              type="submit"
+              disabled={isLoading}
```
quietbits, line 505: We don't need to set `disabled` with `isLoading` state because the `isLoading` prop disables the button internally in the SDS.

image

We should enable the Submit only when all required information is filled out to stay consistent with other forms. For example, the Create New Key modal.

image

PR #402, opened 2025-12-10
SDP-1619: Wire up SEP-24

JiahuiWho:
### What

Use SEP-45 token to initialize SEP-24 for OTP and PII verification.

### Why

Step 2 of SEP-24 verification for embedded wallet

### Known limitations

For local testing:
1. Use single tenant mode
2. Change to `const homeDomain = new URL(API_URL).host` in `authenticateWithSep45.ts`

Reason being:
1. SEP‑45 service only accepts `home_domain` under the base url (localhost:8000) and will bake that home_domain into token.  ([link](https://github.com/stellar/stellar-disbursement-platform-backend/blob/feature/c-accounts/internal/services/sep45_service.go#L814))
2. SEP‑24 query middleware can't extract a tenant from `home_domain` being `localhost:8000` in multi-tenant mode ([link](https://github.com/stellar/stellar-disbursement-platform-backend/blob/feature/c-accounts/internal/sepauth/sep24_auth_middleware.go#L181))

Single-tenant mode lets that middleware skip tenant inference, making the SEP‑45 → SEP‑24 flow work locally.

### Checklist

- [ ] Ready for production


philipliu (commented):
> SEP‑45 service only accepts home_domain under the base url (localhost:8000) and will bake that home_domain into token. ([link](https://github.com/stellar/stellar-disbursement-platform-backend/blob/feature/c-accounts/internal/services/sep45_service.go#L814))

I believe SEP-45 is rejecting non-localhost `home_domain`s because `stellar.toml` is being fetched without the tenant header, so the response falls back to a default `WEB_AUTH_ENDPOINT` like `http://localhost:8000/sep45/auth`.

We should update `getRequiredTomlInfo` to include the tenant header when fetching `stellar.toml`. We should be able to reuse the existing tenant-aware request logic in `fetchApi` for this.

philipliu:
Actually, the fix seems to be more involved. Injecting the tenant name in the request header will still fail the home domain check. We can merge this as is, but fix this in a tenant-aware setup later.

philipliu (approved):

PR #416, opened 2025-12-17
SDP-1953: Implement SEP-24 verification modal

philipliu:
### What

This implements the SEP-24 verification modal.

image

### Why

Update UI with new design

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/src/components/EmbeddedWalletLayout/index.tsx
+++ b/src/components/EmbeddedWalletLayout/index.tsx
@@ -54,7 +54,7 @@ export const EmbeddedWalletLayout = ({
           </Card>
         </div>
       </div>
-      <div id="embedded-wallet-modal-root" />
+      <div id="embedded-wallet-modal-root" className="EmbeddedWalletModalRoot" />
     </div>
   );
 };
--- a/src/components/EmbeddedWalletLayout/index.tsx
+++ b/src/components/EmbeddedWalletLayout/index.tsx
@@ -1,8 +1,9 @@
 import { Card } from "@stellar/design-system";
 import { useEffect, type ReactNode } from "react";
 
-import "./styles.scss";
 import { MODAL_PARENT_ID } from "../EmbeddedWalletModal";
+
+import "./styles.scss";
 
 type EmbeddedWalletLayoutProps = {
   organizationName: string;
--- a/src/components/EmbeddedWalletLayout/index.tsx
+++ b/src/components/EmbeddedWalletLayout/index.tsx
@@ -2,6 +2,7 @@ import { Card } from "@stellar/design-system";
 import { useEffect, type ReactNode } from "react";
 
 import "./styles.scss";
+import { MODAL_PARENT_ID } from "../EmbeddedWalletModal";
 
 type EmbeddedWalletLayoutProps = {
   organizationName: string;
@@ -54,7 +55,7 @@ export const EmbeddedWalletLayout = ({
           </Card>
         </div>
       </div>
-      <div id="embedded-wallet-modal-root" className="EmbeddedWalletModalRoot" />
+      <div id={MODAL_PARENT_ID} className="EmbeddedWalletModalRoot" />
     </div>
   );
 };
--- a/src/components/EmbeddedWalletLayout/index.tsx
+++ b/src/components/EmbeddedWalletLayout/index.tsx
@@ -54,6 +54,7 @@ export const EmbeddedWalletLayout = ({
           </Card>
         </div>
       </div>
+      <div id="embedded-wallet-modal-root" />
     </div>
   );
 };
--- a/src/components/EmbeddedWalletModal/index.tsx
+++ b/src/components/EmbeddedWalletModal/index.tsx
@@ -1,12 +1,15 @@
-import { Button, Modal } from "@stellar/design-system";
+import { Button, Icon, NavButton } from "@stellar/design-system";
+import { useCallback, useEffect } from "react";
+import { createPortal } from "react-dom";
 
 import "./styles.scss";
 
 interface EmbeddedWalletModalProps {
   visible: boolean;
-  onClose: () => void;
+  onClose?: () => void;
+  isDismissible?: boolean;
   title: string;
-  description?: string;
+  description: string;
   contentAlign?: "center" | "left";
   primaryActionLabel?: string;
   onPrimaryAction?: () => void;
@@ -14,9 +17,13 @@ interface EmbeddedWalletModalProps {
   isPrimaryActionDisabled?: boolean;
 }
 
+const MODAL_OPEN_CLASS_NAME = "modal-open";
+const MODAL_PARENT_ID = "embedded-wallet-modal-root";
+
 export const EmbeddedWalletModal = ({
   visible,
   onClose,
+  isDismissible = true,
   title,
   description,
   contentAlign = "center",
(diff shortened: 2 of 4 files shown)


src/components/EmbeddedWalletModal/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+}: EmbeddedWalletModalProps) => {
+  const handleClose = () => {
+    onClose();
+  };
+
+  const contentAlignClass = `EmbeddedWalletModal__content--${contentAlign}`;
+  const headingAlignClass = `EmbeddedWalletModal__heading--${contentAlign}`;
+
+  return (
+    <Modal parentId="embedded-wallet-modal-root" visible={visible} onClose={handleClose}>
```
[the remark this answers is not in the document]
quietbits: Unless we expect to add something extra to `handleClose`, we should use `onClose` here directly.

src/components/EmbeddedWalletModal/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <Modal.Heading>
+        <div className={`EmbeddedWalletModal__heading ${headingAlignClass}`}>
```
quietbits: It might be cleaner to use `data-` attributes instead of classes for styling like this.

```suggestion
        <div className="EmbeddedWalletModal__heading" data-align={contentAlign}>
```

src/components/EmbeddedWalletModal/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <span className="EmbeddedWalletModal__title">{title}</span>
+        </div>
+      </Modal.Heading>
+      <Modal.Body>
+        <div className={`EmbeddedWalletModal__content ${contentAlignClass}`}>
+          {description ? <p className="EmbeddedWalletModal__description">{description}</p> : null}
+        </div>
+      </Modal.Body>
```
quietbits: If there is no `description`, this whole block will be empty. Is that intentional?

src/components/EmbeddedWalletModal/styles.scss
```diff
@@ -0,0 +1,119 @@
+@use "../../styles/styles-utils.scss" as *;
+
+#embedded-wallet-modal-root {
```
quietbits: It's not ideal to use IDs for styling because they make it hard to override styles. Could we make this work with a class?

src/components/EmbeddedWalletModal/styles.scss
```diff
@@ -0,0 +1,119 @@
...
+  --EmbeddedWalletModal-max-width: #{pxToRem(420px)};
+  --EmbeddedWalletModal-padding: #{pxToRem(24px)};
+  --EmbeddedWalletModal-radius: #{pxToRem(8px)};
+  --EmbeddedWalletModal-shadow: 0 #{pxToRem(20px)} #{pxToRem(40px)} rgba(0, 0, 0, 0.16);
+  --EmbeddedWalletModal-overlay: rgba(0, 0, 0, 0.35);
+
+  .Modal {
```
quietbits: We are overriding many `Modal` styles here. Would it be better to make it a standalone component then?

src/components/EmbeddedWalletModal/styles.scss
```diff
@@ -0,0 +1,119 @@
...
+
+.EmbeddedWalletModal__content--center {
+  align-items: center;
+  text-align: center;
+}
+
+.EmbeddedWalletModal__content--left {
+  align-items: flex-start;
+  text-align: left;
+}
```
quietbits: Using `data-align` attribute, this code would become this:
```suggestion
.EmbeddedWalletModal__content {
  display: flex;
  flex-direction: column;
  gap: pxToRem(16px);
  
  &[data-align="center"] {
    align-items: center;
    text-align: center;
  }
  
  &[data-align="left"] {
    align-items: flex-start;
    text-align: left;
  }
}
```

src/components/EmbeddedWalletModal/index.tsx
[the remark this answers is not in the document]
philipliu: It wasn't necessary before, but now it has a null check.

src/components/EmbeddedWalletModal/index.tsx
philipliu: Nope, we should always have a description. Fixed!

src/components/EmbeddedWalletModal/styles.scss
philipliu: Fixed!

src/components/EmbeddedWalletModal/styles.scss
philipliu: Created a new component

src/components/EmbeddedWalletLayout/index.tsx
```diff
@@ -54,6 +54,7 @@ export const EmbeddedWalletLayout = ({
           </Card>
         </div>
       </div>
+      <div id="embedded-wallet-modal-root" className="EmbeddedWalletModalRoot" />
```
quietbits: We could use the const here for the ID `MODAL_PARENT_ID`.

src/components/EmbeddedWalletModal/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+      if (event.key === "Escape") {
+        handleClose();
+      }
+    },
+    [handleClose, isModalDismissible],
+  );
+
+  useEffect(() => {
+    if (!visible || !isModalDismissible) {
+      return undefined;
```
quietbits: I think it will return `undefined` if we don't return anything.
```suggestion
      return;
```

src/components/EmbeddedWalletModal/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+    if (visible) {
+      document.body.classList.add(MODAL_OPEN_CLASS_NAME);
+      return () => {
+        document.body.classList.remove(MODAL_OPEN_CLASS_NAME);
+      };
+    }
+
+    document.body.classList.remove(MODAL_OPEN_CLASS_NAME);
+    return undefined;
+  }, [visible]);
```
quietbits: I'm not quite sure what we're trying to do here. Are we adding and removing a class based on visibility? 

`visible` block shouldn't return anything. The cleanup for `useEffect` should be a function.


```suggestion
  useEffect(() => {
    if (visible) {
      document.body.classList.add(MODAL_OPEN_CLASS_NAME)
    } else {
      document.body.classList.remove(MODAL_OPEN_CLASS_NAME);
    }
    
    return () => {
      document.body.classList.remove(MODAL_OPEN_CLASS_NAME);
    };
  }, [visible]);
```

src/components/EmbeddedWalletModal/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+                title="Close modal"
+                onClick={handleClose}
+                icon={<Icon.XClose />}
+              />
+            </div>
+          ) : null}
+
+          <div className="EmbeddedWalletModal__heading" data-align={contentAlign}>
+            <span className="EmbeddedWalletModal__title">{title}</span>
+          </div>
```
quietbits: We can combine these two elements into a single heading and use `Heading` as `h2`. It also accepts a custom class name for styling.

src/components/EmbeddedWalletModal/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+          <div className="EmbeddedWalletModal__body" data-align={contentAlign}>
+            <p className="EmbeddedWalletModal__description">{description}</p>
+          </div>
```
quietbits, line 95: Same thing here, we can simplify the markup if we only expect to display a paragraph here.

src/components/EmbeddedWalletModal/index.tsx
philipliu: It was supposed to replicate this behaviour that prevents background scrolling while the modal is open, but it seems like it works without it, so I'll remove it https://github.com/stellar/stellar-design-system/blob/ab234b4b903b4b8e65008f7988d9ec2dff435265/%40stellar/design-system/src/components/Modal/index.tsx#L107-L113

src/components/EmbeddedWalletModal/index.tsx
philipliu: I was going to extend this modal later to display icons in the heading. I'll simplify it for now and refactor later if needed.

src/components/EmbeddedWalletLayout/index.tsx
```diff
@@ -2,6 +2,7 @@ import { Card } from "@stellar/design-system";
 import { useEffect, type ReactNode } from "react";
 
 import "./styles.scss";
+import { MODAL_PARENT_ID } from "../EmbeddedWalletModal";
```
quietbits, line 4: Nit: styles should be the last imoprt

quietbits (approved):

PR #418, opened 2025-12-18
SDP-1944: Implement Embedded Wallet profile dropdown and modal

JiahuiWho:
### What

1. store `supportedAssets` and `receiverContact` from the profile API, clear them on logout.

2. new `EmbeddedWalletProfileDropdown` with avatar, contact info, profile/log-out actions.
Screenshot 2025-12-18 at 1 06 07 PM

3. `EmbeddedWalletProfileModal` shows receiver contact, wallet address copy, currency selector, and Stellar Expert link.
Screenshot 2025-12-18 at 1 06 16 PM

### Why

Embedded Wallet Design

### Known limitations

currency selector currently doesn't do anything,  will integrate with balance in next PR

### Checklist

- [ ] Ready for production

### DIFF
--- a/src/components/EmbeddedWalletProfileDropdown/index.tsx
+++ b/src/components/EmbeddedWalletProfileDropdown/index.tsx
@@ -1,0 +1,69 @@
+import { Avatar, Button, Icon } from "@stellar/design-system";
+
+import { DropdownMenu } from "@/components/DropdownMenu";
+import { EmbeddedWalletReceiverContact } from "@/types";
+
+import "./styles.scss";
+
+type Props = {
+  contact: EmbeddedWalletReceiverContact;
+  onOpenProfile: () => void;
+  onLogout: () => void;
+};
+
+const CONTACT_ICON_MAP: Record<string, typeof Icon.UserCircle> = {
+  PHONE_NUMBER: Icon.Phone01,
+  EMAIL: Icon.Mail01,
+};
+
+export const EmbeddedWalletProfileDropdown = ({ contact, onOpenProfile, onLogout }: Props) => {
+  const contactValue = contact.value.trim() || "—";
+  const IconComponent = CONTACT_ICON_MAP[contact?.type?.toUpperCase() || ""] || Icon.UserCircle;
+
+  return (
+    <DropdownMenu
+      triggerEl={
+        <Button variant="tertiary" size="lg" className="WalletHeaderProfileTrigger">
+          <Avatar userName={contactValue} size="sm" />
+          <span className="WalletHeaderProfileTrigger__label">
+            <span className="WalletHeaderProfileTrigger__name">{contactValue}</span>
+            <Icon.ChevronDown className="WalletHeaderProfileTrigger__chevron" />
+          </span>
+        </Button>
+      }
+    >
+      <div className="WalletHeaderProfileMenu" role="menu">
+        <div className="WalletHeaderProfileMenu__info">
+          <IconComponent className="WalletHeaderProfileMenu__infoIcon" />
+          <span className="WalletHeaderProfileMenu__infoValue">{contactValue}</span>
+        </div>
+
+        <div className="WalletHeaderProfileMenu__divider" aria-hidden="true" />
+
+        <Button
+          variant="tertiary"
+          size="md"
+          type="button"
+          className="WalletHeaderProfileMenu__item"
+          onClick={onOpenProfile}
+          role="menuitem"
+        >
+          <Icon.UserCircle className="WalletHeaderProfileMenu__itemIcon" />
+          <span>My profile</span>
+        </Button>
+
+        <Button
+          variant="tertiary"
+          size="md"
+          type="button"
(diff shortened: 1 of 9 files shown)


src/components/EmbeddedWalletProfileDropdown/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+import "./styles.scss";
+
+type Props = {
+  contact: EmbeddedWalletReceiverContact;
+  onOpenProfile: () => void;
+  onLogout: () => void;
+};
+
+const CONTACT_ICON_MAP: Record<string, typeof Icon.UserCircle> = {
```
quietbits: I think the `React.ReactElement` for icon type should work.

```suggestion
const CONTACT_ICON_MAP: Record<string, React.ReactElement> = {
```

src/components/EmbeddedWalletProfileDropdown/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+          variant="tertiary"
+          size="md"
+          type="button"
+          className="WalletHeaderProfileMenu__item WalletHeaderProfileMenu__item--danger"
+          onClick={onLogout}
+          role="menuitem"
+        >
+          <Icon.LogOut01 className="WalletHeaderProfileMenu__itemIcon" />
+          <span>Log out</span>
+        </Button>
```
quietbits: If we use these buttons in multiple places, we could make them components.

src/components/EmbeddedWalletProfileDropdown/styles.scss
```diff
@@ -0,0 +1,120 @@
...
+.WalletHeaderProfileMenu__infoIcon {
+  width: pxToRem(18px);
+  height: pxToRem(18px);
+}
+
+.WalletHeaderProfileMenu__infoValue {
+  word-break: break-all;
+}
+
+.WalletHeaderProfileMenu__item {
```
quietbits, line 86: In cases where we feel like we're overriding too many things, we can create custom components. For example, if a `Button` receives a new style in SDS, we'd need to come here to override it.

src/components/EmbeddedWalletProfileModal/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    const url = `${STELLAR_EXPERT_URL}/account/${contractAddress}`;
+    window.open(url, "_blank", "noopener");
+  };
+
+  return (
+    <Modal visible={isOpen} onClose={onClose}>
+      <Modal.Body>
+        <div className="EmbeddedWalletProfileModal__content">
+          <div className="EmbeddedWalletProfileModal__title">My profile</div>
```
quietbits: We could use `Heading` as `h2` here for better semantics.

src/components/EmbeddedWalletProfileModal/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+            <div className="EmbeddedWalletProfileModal__label">Stellar wallet address</div>
+            <div className="EmbeddedWalletProfileModal__addressRow">
+              <span className="EmbeddedWalletProfileModal__copyText">{formattedAddress}</span>
+              {contractAddress ? (
+                <div className="EmbeddedWalletProfileModal__copyAction">
+                  <CopyWithIcon textToCopy={contractAddress} iconSizeRem="1.25" doneLabel="Copied">
+                    <span />
+                  </CopyWithIcon>
+                </div>
+              ) : null}
```
quietbits, line 55: What if we render this conditionally:
1. If there is a contract address, wrap it in `CopyWithIcon` component
2. If there is no address, return only the text

The current approach has too many elements that are not needed, which makes it harder to follow what is going on.

quietbits (approved):

src/components/EmbeddedWalletProfileModal/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+              value={currency}
+              onChange={(event) => setCurrency(event.target.value)}
+            >
+              {currencies.map((currency) => (
+                <option key={currency} value={currency}>
+                  {currency}
+                </option>
+              ))}
+            </Select>
+          </div>
```
philipliu: We discussed removing the current selector since we won't be showing prices in the initial release.

philipliu (approved):

src/components/EmbeddedWalletProfileModal/index.tsx
JiahuiWho: Wrapping the address text inside `CopyWithIcon` turns the entire row into the CopyText trigger, so the “Copied” tooltip anchors to the right edge of that full-width element. With SDS’s default `tooltipPlacement="right"`, the bubble extends past the modal boundary, forcing the container to grow (and the content becomes horizontally scrollable). If we try to hide the overflow to stop the scroll, the tooltip gets clipped instead.
Screenshot 2025-12-22 at 12 50 10 PM

For now I would keeping the text and CopyWithIcon as separate siblings limits the trigger to the icon area, so the tooltip stays near the icon and never affects layout.

PR #425, opened 2025-12-24
Fix: Logout user when wallet JWT expires

philipliu:
### What

This adds session token expiry checks to the embedded wallet API helpers so that the user is logged out when the token expires.

### Why

Previously, users stayed logged in despite their JWTs expiring.

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/src/helpers/fetchWalletApi.ts
+++ b/src/helpers/fetchWalletApi.ts
@@ -1,7 +1,10 @@
+import { differenceInMinutes, fromUnixTime } from "date-fns";
+
 import { handleApiResponse } from "@/api/handleApiResponse";
-import { SESSION_EXPIRED, SESSION_EXPIRED_EVENT } from "@/constants/settings";
+import { SESSION_EXPIRED, WALLET_SESSION_EXPIRED_EVENT } from "@/constants/settings";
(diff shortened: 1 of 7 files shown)


src/helpers/fetchWalletApi.ts
```diff
@@ -10,7 +13,15 @@ export const fetchWalletApi = async <ResponseType>(
   const token = localStorageWalletSessionToken.get();
 
   if (!token) {
-    document.dispatchEvent(new CustomEvent(SESSION_EXPIRED_EVENT));
+    document.dispatchEvent(new CustomEvent(WALLET_SESSION_EXPIRED_EVENT));
+    throw new Error(SESSION_EXPIRED);
```
quietbits: Should we create a new variable here to separate it from the global session expired?

quietbits (approved):

src/helpers/fetchWalletApi.ts
philipliu: Good call. Addressed in https://github.com/stellar/stellar-disbursement-platform-frontend/pull/425/commits/02fd57b46eee2fba84e2355fd5e4b3d52816dda6

PR #432, opened 2026-01-08
SDP-1634: Hide Customize Invite Message for KWA 

JiahuiWho:
### What

1. conditionally render `DisbursementInviteMessage` only when the registration contact type is **NOT** a KWA
2. Suppress the `set-state-in-effect` lint error for now, tarcking in https://stellarorg.atlassian.net/browse/SDP-1969

### Why

```
 error  Error: Calling setState synchronously within an effect can trigger cascading renders
[1] 
[1] Effects are intended to synchronize state between React and external systems such as manually updating the DOM, state management libraries, or other platform APIs. In general, the body of an effect should do one or both of the following:
[1] * Update external systems with the latest state from React.
[1] * Subscribe for updates from some external system, calling setState in a callback function when external state changes.
[1] 
[1] Calling setState synchronously within an effect body causes cascading renders that can hurt performance, and is not recommended. (https://react.dev/learn/you-might-not-need-an-effect).
```
### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/src/pages/DisbursementsNew.tsx
+++ b/src/pages/DisbursementsNew.tsx
@@ -1,9 +1,25 @@
-import { useEffect, useRef, useState } from "react";
 import { Badge, Card, Heading } from "@stellar/design-system";
+import { BigNumber } from "bignumber.js";
+import { useEffect, useRef, useState } from "react";
 import { useDispatch } from "react-redux";
 import { useNavigate } from "react-router-dom";
-import { BigNumber } from "bignumber.js";
 
+import { AccountBalances } from "@/components/AccountBalances";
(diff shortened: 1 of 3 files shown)


marwen-abid:
@JiahuiWho Please update CHANGELOG.md

src/pages/DisbursementsNew.tsx
```diff
@@ -112,10 +117,6 @@ export const DisbursementsNew = () => {
     }
   };
 
-  const handleScrollToTop = () => {
```
JiahuiWho, line 115: Move to the top to avoid "use but not declared" error at line 64

quietbits (approved):

PR #438, opened 2026-01-14
SDP-1943: Implement remaining transfer flow

philipliu:
### What

This implements the remaining transfer flow.

### XLM Transfer

https://github.com/user-attachments/assets/f049c21a-bfd9-472d-80bd-9ff641d90946

### USDC Transfer

https://github.com/user-attachments/assets/f86a6484-55df-40ec-8ffa-237d66de4a10

### Transfer failed during simulation

image

### Why

Use the new design

### Known limitations

This does not implement the transfer detail validation or the amount text resizing since the PR is already so large. Those will be implemented in a follow-up commit.

### Checklist

- [ ] Ready for production


JiahuiWho:
LGTM. 

Would it make sense to keep the entire transfer flow inside a single modal container? With internal step views so the focus state stays stable.

src/apiQueries/useSendWalletPayment.ts
```diff
@@ -36,10 +39,43 @@ interface UseSendWalletPaymentOptions {
...
   balance: string;
+  assetCode: string;
+  assetIssuer?: string | null;
   onSuccess?: (result: SendWalletPaymentResult) => void | Promise<void>;
+  onSigned?: () => void | Promise<void>;
 }
 
 const SIGNATURE_EXPIRATION_LEDGER_BUFFER = 10;
+export const SIMULATION_ERROR_CODE = "SIMULATION_FAILED";
+export const TRANSACTION_FAILED_CODE = "TRANSACTION_FAILED";
```
quietbits: Should we move these to `/constants/settings.ts` with more descriptive names?

src/apiQueries/useSendWalletPayment.ts
```diff
@@ -159,15 +212,29 @@ export const useSendWalletPayment = ({
...
+      try {
+        simulationResult = await simulateTransferOperation({
+          operation: transferOperation,
+          networkPassphrase,
+        });
+      } catch (error) {
+        if (
+          error &&
+          typeof error === "object" &&
+          "extras" in error &&
```
quietbits: We use this check in a couple of places. Should we create a helper?

src/components/EmbeddedWalletAssetLogo/styles.scss
```diff
@@ -0,0 +1,62 @@
...
+
+  &[data-size="md"] {
+    width: pxToRem(32px);
+    height: pxToRem(32px);
+  }
+
+  &[data-size="lg"] {
+    width: pxToRem(40px);
+    height: pxToRem(40px);
+  }
```
quietbits, line 36: We could use a CSS variable to set and change the size (width and height are the same, so we only need one variable).

```suggestion
.EmbeddedWalletAssetLogo {
  --EmbeddedWalletAssetLogo-size: #{pxToRem(20px)};

  display: inline-flex;
  align-items: center;
  justify-content: center;
  border-radius: 50%;
  overflow: hidden;
  border: 2px solid var(--sds-clr-gray-06);
  background: var(--sds-clr-gray-01);
  flex-shrink: 0;
  box-sizing: border-box;
  width: var( --EmbeddedWalletAssetLogo-size);
  height: var( --EmbeddedWalletAssetLogo-size);

  &[data-size="sm"] {
     --EmbeddedWalletAssetLogo-size: #{pxToRem(20px)};
  }

  &[data-size="md"] {
     --EmbeddedWalletAssetLogo-size: #{pxToRem(32px)};
  }

  &[data-size="lg"] {
     --EmbeddedWalletAssetLogo-size: #{pxToRem(40px)};
  }
```

src/components/EmbeddedWalletAssetLogo/styles.scss
```diff
@@ -0,0 +1,62 @@
...
+
+  &[data-size="md"] .Avatar__initials {
+    font-size: pxToRem(14px);
+    transform: translateY(1px);
+  }
+
+  &[data-size="lg"] .Avatar__initials {
+    font-size: pxToRem(18px);
+    transform: translateY(1px);
+  }
```
quietbits: Same here.

src/components/EmbeddedWalletReviewTransferModal/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+          <div className="EmbeddedWalletReviewTransferModal__details">
+            <div className="EmbeddedWalletReviewTransferModal__detail">
+              <div className="EmbeddedWalletReviewTransferModal__detailLabel">From (sender)</div>
+              <div className="EmbeddedWalletReviewTransferModal__detailRow">
+                <span className="EmbeddedWalletReviewTransferModal__address">{senderDisplay}</span>
+                {trimmedSender ? (
+                  <div className="EmbeddedWalletReviewTransferModal__copyAction">
+                    <CopyWithIcon textToCopy={trimmedSender} iconSizeRem="1.1" doneLabel="Copied">
+                      <span className="EmbeddedWalletReviewTransferModal__copySpacer" />
```
quietbits: What is the spacer used for?

src/components/EmbeddedWalletReviewTransferModal/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+                <span className="EmbeddedWalletReviewTransferModal__assetBadge">
+                  {xlmLogo ? <img src={xlmLogo} alt="XLM icon" /> : <Icon.Coins03 />}
+                </span>
+              ) : null}
+            </div>
+            <div className="EmbeddedWalletReviewTransferModal__summaryLabel">You're sending</div>
+            <div className="EmbeddedWalletReviewTransferModal__summaryAmount">{displayAmount}</div>
+          </div>
+
+          <div className="EmbeddedWalletReviewTransferModal__details">
```
quietbits, line 96: Receiver and sender blocks seem the same. We could create this as a local component to use for both.

quietbits (approved):
PR looks good, just left some minor suggestions.

To @JiahuiWho's point, having multiple modals is quite complex, but moving all the logic to a single modal might be more complex because we're dealing with different statuses and states. 🤔 Let's leave this as is. As long as we have only one modal open at a time (which we should because we have a single step property), we should be alright.

src/components/EmbeddedWalletReviewTransferModal/index.tsx
philipliu: It's not needed. It was added originally to center the icon within the border, but it looks like it also works without it.

PR #452, opened 2026-01-23
SDP-1998: RPC client allowHttp based on API_URL protocol instead of NODE_ENV

marwen-abid:
### What

Changed allowHttp configuration in createAuthenticatedRpcServer.ts to be determined by the API_URL protocol instead of NODE_ENV.

### Why

The embedded wallet verification flow was silently failing in production Docker builds:

1. Production builds set NODE_ENV=production, causing allowHttp=false
2. When API_URL is HTTP (e.g., http://localhost:8000), the Stellar SDK's rpc.Server silently rejects requests
3. This blocked RPC calls (getNetwork(), getLatestLedger()) needed before the passkey prompt, causing the "Start verification" button to load indefinitely

The fix respects the configured API_URL protocol—if the developer explicitly configures HTTP, the RPC client should allow it. This is consistent with all other API calls which already use the same API_URL.

### Checklist

- [ ] Ready for production

### DIFF
--- a/src/helpers/createAuthenticatedRpcServer.ts
+++ b/src/helpers/createAuthenticatedRpcServer.ts
@@ -1,6 +1,7 @@
 import { rpc } from "@stellar/stellar-sdk";
 
 import { API_URL } from "@/constants/envVariables";
+
 import { getSdpTenantName } from "@/helpers/getSdpTenantName";
 import { localStorageSessionToken } from "@/helpers/localStorageSessionToken";
 import { localStorageWalletSessionToken } from "@/helpers/localStorageWalletSessionToken";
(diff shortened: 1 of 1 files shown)


src/helpers/createAuthenticatedRpcServer.ts
```diff
@@ -20,9 +21,9 @@ export const createAuthenticatedRpcServer = (
 
   const rpcProxyUrl = `${API_URL}/rpc/${authType}`;
 
-  const isDevelopment = process.env.NODE_ENV === "development";
```
marwen-abid, line 23: The issue was that all docker builds were "production builds" -> NODE_ENV always "production" -> this was always requiring `https`, even when the API_URL is http (`http://localhost:8000`)

philipliu (approved):

PR #504, opened 2026-03-16
Reports Feature: Add Reports page for statement export

rodrigogsn:
### What

- **Reports feature (first release)**: Adds a new Reports page and report PDF flows. Users can download a wallet statement PDF and a transaction notice PDF; all report requests go to the new `/reports` API.
- **New Reports page**: Route `/reports` with two sections — **Wallet Statement** (date range, Download Statement → statement PDF) and **Individual Transaction Notice** (search by Transaction ID or Payment ID, select payment, optional notes, Download Notice → transaction notice PDF). Nav entry and route are gated by the Reports feature flag.
- **New API usage**:
  - **Statement PDF**: `useStatementExport` calls **GET `/reports/statement`** with query params `from_date`, `to_date`, optional `asset_code`, optional `base_url`.
  - **Transaction notice PDF**: `useTransactionNoticeExport` calls **GET `/reports/payment/{paymentId}`** with optional `internal_notes`, optional `base_url`.
- **Feature flag for Reports**: The Reports menu item and `/reports` route are controlled by **`REACT_APP_ENABLE_REPORTS_FEATURE`** (build-time) or **`ENABLE_REPORTS_FEATURE`** (runtime via `window._env_`). The flag supports three modes:
  - `"true"` — Reports is enabled for **all** organizations.
  - `"false"` or empty/unset — Reports is **disabled** for all organizations.
  - Pipe-separated organization names (e.g. `"Org A|Org B"`) — Reports is enabled **only** for the listed organizations (case-insensitive, whitespace-trimmed).

  Read in `src/constants/envVariables.ts` as `REPORTS_FEATURE_ORGANIZATION_NAMES`; evaluated by `isReportsFeatureEnabledForOrganization()` against the current organization name. Used in `src/App.tsx` (route gate) and `src/components/InnerPage/index.tsx` (sidebar nav item). Typings in `src/vite-env.d.ts`.

  Examples:
  ```
  REACT_APP_ENABLE_REPORTS_FEATURE=true            # all orgs
  REACT_APP_ENABLE_REPORTS_FEATURE="Demo Org|Acme" # only these orgs
  REACT_APP_ENABLE_REPORTS_FEATURE=false            # disabled
  ```
- **Table component fix**: Resolved "Cannot create components during render" in `Table` by rendering the sort icon as a JSX value (`sortIcon`) instead of an inline component, avoiding remounts.

### Why

- Ship the first release of the Reports feature so users can export wallet statement and transaction notice PDFs from the UI.
- Use a feature flag so Reports can be enabled globally, disabled, or selectively enabled per organization name — without code changes — to support staged rollouts.

### Known limitations

- N/A (or: Reports is only visible when the feature flag is enabled; backend must expose GET `/reports/statement` and GET `/reports/payment/{id}`.)

### Checklist

- [ ] Ready for production


src/apiQueries/useTransactionNoticeExport.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export const INTERNAL_NOTES_MAX_LENGTH = 500;
+
+export type TransactionNoticeExportParams = {
+  paymentId: string;
+  internalNotes?: string;
+  baseUrl?: string;
+};
+
+function getFilenameFromContentDisposition(header: string | null, fallback: string): string {
```
JiahuiWho: Is this something we can extract to shared helper?

src/components/TransactionNoticeCard/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+    const notesTrimmed = internalNotes.trim();
+    const internalNotesParam =
+      notesTrimmed.length > INTERNAL_NOTES_MAX_LENGTH
+        ? notesTrimmed.slice(0, INTERNAL_NOTES_MAX_LENGTH)
+        : notesTrimmed || undefined;
+    exportTransactionNotice({
+      paymentId: selectedId,
+      internalNotes: internalNotesParam,
+      baseUrl: organization.data.baseUrl ?? undefined,
+    });
```
JiahuiWho, line 72: The notes are truncated here and again in `exportTransactionNotice()`. Is this intended?

src/apiQueries/useStatementExport.ts
```diff
@@ -0,0 +1,89 @@
...
+    resolve();
+  } catch (e) {
+    reject(e);
+  }
+}
+
+export const useStatementExport = () => {
+  const mutation = useMutation<void, AppError, StatementQueryParams>({
+    mutationFn: (params: StatementQueryParams): Promise<void> =>
+      new Promise<void>((resolve, reject) => {
```
JiahuiWho: This is the explicit Promise construction anti-pattern because `fetchApi` already returns a Promise. Wrapping it in `new Promise()` just to thread resolve/reject through a callback adds unnecessary complexity.
You can follow the existing `getExport` pattern to handle it.

src/apiQueries/useTransactionNoticeExport.ts
```diff
@@ -0,0 +1,101 @@
...
+    });
+    resolve();
+  } catch (e) {
+    reject(e);
+  }
+}
+
+export const useTransactionNoticeExport = () => {
+  const mutation = useMutation<void, AppError, TransactionNoticeExportParams>({
+    mutationFn: (params: TransactionNoticeExportParams): Promise<void> =>
```
JiahuiWho: Same anti-pattern issue here

quietbits (commented):
I will do a deeper review once @JiahuiWho 's feedback is addressed.

src/App.tsx
```diff
@@ -311,6 +327,8 @@ export const App = () => {
                 </PrivateRoute>
               }
             />
+            {/* Reports */}
+            <Route path={Routes.REPORTS} element={<ReportsPageGate />} />
```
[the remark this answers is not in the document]
rodrigogsn: Created REPORTS entry in Routes enum (it was accidentally removed when taking out CKL CI files)

src/components/WalletStatementCard/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+import { useStatementExport } from "@/apiQueries/useStatementExport";
+import { ErrorWithExtras } from "@/components/ErrorWithExtras";
+import { InfoTooltip } from "@/components/InfoTooltip";
+import {
+  getThisMonthDates,
+  getLastMonthDates,
+  getQTDDates,
+  getYTDDates,
+} from "@/helpers/getStatementPeriodDates";
```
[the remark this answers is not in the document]
rodrigogsn: Recreated date format functions (they were accidentally removed when taking out CKL CI files)

src/components/InnerPage/index.tsx
```diff
@@ -84,6 +85,16 @@ export const InnerPage = ({ children, isNarrow, isCardLayout }: InnerPageProps)
       route: Routes.ANALYTICS,
       icon: <Icon.LineChartUp01 />,
     },
+    ...(reportsEnabled
+      ? [
+          {
+            id: "nav-reports",
+            label: "Reports",
+            route: Routes.REPORTS,
```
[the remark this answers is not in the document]
rodrigogsn: Created REPORTS entry in Routes enum (it was accidentally removed when taking out CKL CI files)

src/apiQueries/useTransactionNoticeExport.ts
rodrigogsn: Sure, thanks for pointing that out 🚀

src/apiQueries/useStatementExport.ts
rodrigogsn: Thanks, you are right. Just refactored to use solely fetchApi promises.

src/apiQueries/useTransactionNoticeExport.ts
rodrigogsn: Same solution as above.

src/constants/envVariables.ts
```diff
@@ -71,9 +74,28 @@ export const {
...
+  const raw = REPORTS_FEATURE_ORGANIZATION_NAMES;
+  if (!raw || raw.trim().toLowerCase() === "false") return false;
+  if (raw.trim().toLowerCase() === "true") return true;
+  const allowed = new Set(
+    raw
+      .split("|")
+      .map((s) => s.trim().toLowerCase())
+      .filter(Boolean),
+  );
+  return allowed.has(organizationName.trim().toLowerCase());
```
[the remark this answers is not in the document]
rodrigogsn: PR description updated

marwen-abid:
@quietbits I think this PR is ready for your review. CKL is done addressing @JiahuiWho's review.

src/apiQueries/useStatementExport.ts
```diff
@@ -0,0 +1,61 @@
...
+import { getFilenameFromContentDisposition } from "@/helpers/getFilenameFromContentDisposition";
+import { normalizeApiError } from "@/helpers/normalizeApiError";
+import { saveFile } from "@/helpers/saveFile";
+import { AppError, StatementQueryParams } from "@/types";
+
+function statementFilename(fromDate: string, toDate: string): string {
+  const from = fromDate.replaceAll("-", "");
+  const to = toDate.replaceAll("-", "");
+  return `statement_${from}-${to}.pdf`;
+}
```
quietbits: Nit: we prefer `const + arrow function` instead of `function`, if we want to stay consistent.

src/components/EmptyStateMessage/EmptyStateMessage.tsx
```diff
@@ -4,12 +4,30 @@ import "./styles.scss";
...
+        <>
+          {title && <span className="EmptyStateMessage__title">{title}</span>}
+          {description && <span className="EmptyStateMessage__description">{description}</span>}
+        </>
+      ) : (
+        message
+      )}
+    </div>
+  );
+};
```
quietbits, line 40: If `description` is provided without `title`, we don't want to show it?

It is also possible that there will be no `message` or `icon`, so we might want to add another check. Overall, we might want to add better types using TypeScript discriminated union, because in the current state it's possible to render this component without any props. Not sure if that is the intent here.

src/components/EmptyStateMessage/styles.scss
```diff
@@ -4,4 +4,21 @@
   justify-content: center;
   gap: 4px;
   color: var(--sds-clr-gray-09);
+
+  &--stacked {
+    flex-direction: column;
+    gap: 8px;
```
quietbits: For any `px` value greater than 1, we should use `pxToRem(8px)` helper method. Sorry that we don't have a proper guideline for these things.

src/components/EmptyStateMessage/styles.scss
```diff
@@ -4,4 +4,21 @@
...
+    text-align: center;
+
+    .EmptyStateMessage__title {
+      font-weight: 600;
```
quietbits: We should use design system variables where possible. The best way to view what's available is inspecting `root` node.

```suggestion
      font-weight: var(--sds-fw-semi-bold);
```

src/components/EmptyStateMessage/styles.scss
```diff
@@ -4,4 +4,21 @@
...
+      color: var(--sds-clr-gray-09);
+    }
+
+    .EmptyStateMessage__description {
+      font-weight: 400;
+      color: var(--sds-clr-gray-09);
+      font-size: 0.875rem;
+    }
```
quietbits, line 24: Same here, we should use SDS (Stellar Design System) variables for weight and `pxToRem()` helper for size. You might also look at `Text` component which provides props to set basic styles.

src/components/Table/index.tsx
```diff
@@ -63,7 +62,7 @@ const HeaderCell: React.FC<HeaderCellProps> = ({
     >
       {elementLeft || sortDirection ? (
         <span className="Table-v2__header__cell" {...sortButtonProps}>
-          {elementLeft ?? null} {children} <SortIconEl />
+          {elementLeft ?? null} {children} {sortIcon}
```
quietbits, line 65: Curious, why this change? Does it not like component approach anymore?

quietbits (commented):
Thanks for the PR @rodrigogsn ! 🎉 

I left a few comments, mostly related to keeping the code base consistent following our (undocumented 😊) conventions. Sorry that we don't have a proper doc for those (maybe one day 🤞)! It's mostly about using SDS (Stellar Design System) CSS variables where possible and `pxToRem()` helper for sizes (we found that using `px` from design is more accurate this way). Let me know if you have any questions!

README.md
```diff
@@ -116,6 +116,10 @@ Set the value to true to avoid fetching the file if the Window ENV is not used.
 > [!NOTE]
 > Set `REACT_APP_DISABLE_TENANT_PREFIL_FROM_DOMAIN=true` to skip prefilling the hostname from the domain.
 
+<!-- prettier-ignore -->
+> [!NOTE]
+> Set `REACT_APP_ENABLE_REPORTS_FEATURE` to control the Reports feature (menu item and `/reports` route). Use a pipe-separated list of organization names (case-insensitive); the pipe character `|` is the separator (spaces around pipes are optional), and names may contain spaces inside them, e.g. `"Demo Organization | Other Org"`. Use `true` to enable for all organizations, or `false` (or leave unset) to disable for everyone.
```
marwen-abid: This is not very flexible and it's pretty error-prone.  We should probably have a flag in the `organizations` table for `reporting_enabled`.  Similar to the current `captcha_disabled` or `mfa_disabled`. 

That configuration can then be loaded via the `/app-config` endpoint, which loads tenant specific configuration. 

It currently returns the following : 
```
{
  "captcha_type": "GOOGLE_RECAPTCHA_V3",
  "captcha_site_key": "",
  "captcha_disabled": true
}
``` 
https://github.com/stellar/stellar-disbursement-platform-backend/blob/develop/internal/serve/httphandler/app_config_handler.go

src/components/WalletStatementCard/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+];
+
+const PERIOD_GETTERS: Record<StatementPeriodKey, () => { fromDate: string; toDate: string }> = {
+  this_month: getThisMonthDates,
+  last_month: getLastMonthDates,
+  qtd: getQTDDates,
+  ytd: getYTDDates,
+};
+
+export const WalletStatementCard = () => {
```
JiahuiWho, line 41: The date inputs in the Wallet Statement card (and potentially the Transaction Notice card) allow selecting future dates within the current month. Consider adding a `max` constraint (e.g. today's date) on the date inputs to prevent users from selecting dates in the future.

src/components/EmptyStateMessage/EmptyStateMessage.tsx
rodrigogsn: Thank you, you are right. Just created two type variants to allow either message or title+description. Please take a look and let me know if this is a more solid approach

src/components/EmptyStateMessage/styles.scss
rodrigogsn: Done. I've also fixed the gap from parent class EmptyStateMessage using the same logic 😄

src/components/Table/index.tsx
rodrigogsn: It's throwing a `eslint(react-hooks/static-components)` error

_Components created during render will reset their state each time they are created. Declare components outside of render._

Do you believe the approach can be improved?

src/components/WalletStatementCard/index.tsx
rodrigogsn: Indeed, thank you

src/components/EmptyStateMessage/EmptyStateMessage.tsx
quietbits: Nice, thank you!

src/components/WalletStatementCard/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+        <div className="WalletStatementCard__actions">
+          <Button
+            size="md"
+            variant="secondary"
+            icon={<Icon.Download01 />}
+            iconPosition="left"
+            onClick={handleDownload}
+            disabled={!isValidRange || isPending}
+            isLoading={isPending}
```
quietbits: Nit: `isLoading` will show a spinner and the button will be disabled. We don't need to set `disabled` as well.

```suggestion
            disabled={!isValidRange}
            isLoading={isPending}
```

src/components/Table/index.tsx
quietbits: No, this is fine. I was just curious, because we get new ESLint rules with newer versions. Thanks for updating!

src/components/TransactionNoticeCard/styles.scss
```diff
@@ -0,0 +1,71 @@
+@use "../../styles/styles-utils.scss" as *;
+
+.TransactionNoticeCard {
+  &__search {
+    margin-bottom: 1rem;
```
quietbits: 👮‍♀️ Sorry for being picky about these, but we shouldn't use `rem` directly in the CSS files. We should use `pxToRem` everywhere instead (when value is greater than 1px, if smaller use px). There are quite a few `rem` in this file.

```suggestion
    margin-bottom: pxToRem(16px);
```

src/components/WalletStatementCard/styles.scss
```diff
@@ -0,0 +1,33 @@
+.WalletStatementCard {
+  &__period {
+    margin-bottom: 1rem;
```
quietbits: A few of these in this file as well.

```suggestion
    margin-bottom: pxToRem(16px);
```

README.md
rodrigogsn: Thank you, above changes were made in both frontend and backend regarding `/app-config.` Please take a look

src/components/TransactionNoticeCard/styles.scss
rodrigogsn: Thanks for noticing!

src/components/WalletStatementCard/index.tsx
rodrigogsn: Absolutely, thank you

quietbits (approved):

PR #539, opened 2026-05-27
SDP-2092 chore: validate soroban auth entry invocation in useSendWalletPayment

hypekostas:
### What

- Adds `validateTransferAuthEntries` to `useSendWalletPayment.ts`, which validates the Soroban authorization entry invocation returned by the RPC simulation before it is passed to the user's passkey for signing.

### Why

- In response to previous SDP-2090: The SDP embedded wallet uses WebAuthn passkeys to authorize Soroban transfers, with the authorization entry constructed from the RPC simulation response. Currently, client is blind-signing whatever the simulation returns.
- As a code hardening measure, the client now validates that the auth entry returned matches the transfer the user initiated before passkey prompt fires, verifying the token contract, function name, sender, recipient, and amount. 
- This mirrors the existing `validateAuthorizationEntry` pattern already in place for the SEP-45 authentication flow (`src/helpers/sep45Authentication/sign.ts`).

### Assumptions
- An honest SAC transfer simulation will return exactly one auth entry of address-credential type

### Design decisions
`authEntries.length !== 1` guard:
- Ensures that the honest RPC simulation only returned exactly one auth entry.
- Previous check for truthiness of `authEntries.length` now obsolete and removed.

Expanded try-catch block in `mutationFn`:`useSendWalletPayment`:
- `simulateTransferOperation`, and `validateTransferAuthEntries` all belong to the same phase (validate simulation response before signing), so they share one try-catch block. 
- Errors not explicitly handled by `validateTransferAuthEntries` (e.g. malformed ScVal type in args throwing a raw XDR error) are therefore still caught as simulationErrors.

Contract address credentials check extracted to shared helper `isAddressCredentialForContract.ts`:
- The logic validating address credentials required by `validateTransferAuthEntries` is identical to that in `signSorobanAuthorization.ts/shouldSignEntry`, therefore it is extracted to a shared helper to avoid repetition.

A note on testing:
- As the frontend repo contains no tests, the validator function was tested using a Claude-generated Node test script that ran an inline copy of the function with the help of the `@stellar/stellar-sdk`. 

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/src/apiQueries/useSendWalletPayment.ts
+++ b/src/apiQueries/useSendWalletPayment.ts
@@ -177,17 +177,16 @@ const validateTransferAuthEntries = (
   destination: string,
   amountInStroops: bigint,
 ) => {
-  const userEntries = authEntries.filter((entry) => {
-    return isAddressCredentialForContract(entry, contractAddress);
-  });
-
-  if (userEntries.length !== 1) {
-    throw createSimulationError(
-      `Expected exactly 1 auth entry for user wallet, got ${userEntries.length}`,
-    );
+  if (authEntries.length !== 1) {
+    throw createSimulationError(`Expected exactly 1 auth entry, got ${authEntries.length}`);
   }
 
-  const invocation = userEntries[0].rootInvocation();
+  const entry = authEntries[0];
+  if (!isAddressCredentialForContract(entry, contractAddress)) {
+    throw createSimulationError("Auth entry is not for expected user wallet contract");
+  }
+
+  const invocation = entry.rootInvocation();
   if (invocation.subInvocations().length) {
     throw createSimulationError(
       "Auth entry invocation authorizes sub-invocation(s) to another contract",
@@ -296,10 +295,6 @@ export const useSendWalletPayment = ({
         });
 
         authEntries = simulationResult.result?.auth ?? [];
-
-        if (!authEntries.length) {
-          throw createSimulationError("Simulation did not return any authorization entries");
-        }
 
         validateTransferAuthEntries(
           authEntries,
--- a/src/apiQueries/useSendWalletPayment.ts
+++ b/src/apiQueries/useSendWalletPayment.ts
@@ -12,6 +12,7 @@ import {
   TransactionBuilder,
   nativeToScVal,
   rpc,
+  scValToBigInt,
   xdr,
 } from "@stellar/stellar-sdk";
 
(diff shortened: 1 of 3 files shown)


src/apiQueries/useSendWalletPayment.ts
```diff
@@ -168,6 +169,87 @@ const simulateTransferOperation = async ({
...
+    const address = addressCredentials.address();
+
+    if (address.switch() !== xdr.ScAddressType.scAddressTypeContract()) {
+      return false;
+    }
+
+    return Address.fromScAddress(address).toString() === contractAddress;
+  });
+
+  if (userEntries.length !== 1) {
```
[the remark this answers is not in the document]
hypekostas: @marwen-abid Does the assumption that an honest transfer simulation will return one auth entry stand? If so, I can change the `if (userEntries.length !== 1)` check to a `if (authEntries.length !== 1)` check at the start of the function. This way, we check that one entry is always returned, rather than checking that only one entry (of many) matches the user's wallet.

src/apiQueries/useSendWalletPayment.ts
```diff
@@ -168,6 +169,87 @@ const simulateTransferOperation = async ({
...
 };
 
+const validateTransferAuthEntries = (
+  authEntries: xdr.SorobanAuthorizationEntry[],
+  contractAddress: string,
+  assetContractId: string,
+  destination: string,
+  amountInStroops: bigint,
+) => {
+  const userEntries = authEntries.filter((entry) => {
```
marwen-abid: We should exact this to a shared helper as it duplicates `shouldSignEntry`  in `signSorobanAuthorization.ts` as you mentioned.

src/apiQueries/useSendWalletPayment.ts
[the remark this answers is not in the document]
marwen-abid: Given that we're being defensive, we should address this yes.

src/apiQueries/useSendWalletPayment.ts
```diff
@@ -168,6 +169,87 @@ const simulateTransferOperation = async ({
...
+  }
+
+  const invocation = userEntries[0].rootInvocation();
+  if (invocation.subInvocations().length) {
+    throw createSimulationError(
+      "Auth entry invocation authorizes sub-invocation(s) to another contract",
+    );
+  }
+
+  const contractFunction = invocation.function().contractFn();
```
marwen-abid, line 203: nit: This would still get caught with the catch but it would produce an obscure error. Maybe worth checking: 
```js
if (
    authorizedFunction.switch() !==
    xdr.SorobanAuthorizedFunctionType.sorobanAuthorizedFunctionTypeContractFn()
  ) {
    throw createSimulationError(
      "Auth entry does not authorize a contract function invocation",
    );
  }
```

src/apiQueries/useSendWalletPayment.ts
hypekostas: Addressed!

src/apiQueries/useSendWalletPayment.ts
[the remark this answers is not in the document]
hypekostas: Addressed!

src/apiQueries/useSendWalletPayment.ts
hypekostas: Addressed, thanks for catching this!

quietbits (approved):

marwen-abid (approved):
LGTM! 🚀

PR #556, opened 2026-07-16
Feature

Shegitu:
...create a frontend demo


hypekostas:
This PR applies Sapcone branding and enables a mocked-backend demo mode by default, which is not a change we can accept. It is apparent that the PR is opened against upstream by mistake, so it will be closed. Thank you for your interest in the SDP!

PR #582, opened 2026-09-01
SDP-2137 chore: add DistributionAccountLabel to WalletBalancesOverview

hypekostas:
### What

Cosmetic enhancement: distribution account label (which includes color coded dot and DEFAULT badge) added to the `WalletBalancesOverview` Home component. This restores consistency with the distribution account color coding applied in the account switcher and across various tables in the dashboard.

Before: 
Before

After: 
image

### Why

The switcher already gives every account a stable color and a `DEFAULT` badge, but the Home overview listed accounts as plain names. Users had to read the name to match a row to the account they'd selected. This makes the two surfaces read the same way.

### Known limitations

- Two other places still render the default marker as literal `(default)` text: the single-account bar (`ActiveWalletBar/index.tsx`) and the account picker in the disbursement wizard (`DisbursementsNew.tsx`).

### Checklist

- [ ] Ready for production

### DIFF
--- a/src/components/WalletBalancesOverview.tsx
+++ b/src/components/WalletBalancesOverview.tsx
@@ -5,6 +5,7 @@ import { Button, Card, Notification } from "@stellar/design-system";
 import { AddDistributionWalletModal } from "@/components/AddDistributionWalletModal";
 import { ArchiveDistributionWalletModal } from "@/components/ArchiveDistributionWalletModal";
 import { AssetAmount } from "@/components/AssetAmount";
+import { DistributionAccountLabel } from "@/components/DistributionAccountLabel";
 import { InfoTooltip } from "@/components/InfoTooltip";
 import { ManageWalletAccessModal } from "@/components/ManageWalletAccessModal";
 import { ShowForRoles } from "@/components/ShowForRoles";
@@ -12,7 +13,6 @@ import { ShowForRoles } from "@/components/ShowForRoles";
 import { useDistributionWalletBalance } from "@/apiQueries/useDistributionWalletBalance";
 import { useDistributionWallets } from "@/apiQueries/useDistributionWallets";
 
-import { distributionAccountDisplayName } from "@/helpers/distributionAccountDisplayName";
 import { parseAssetKey } from "@/helpers/parseAssetKey";
 
 import { useIsUserRoleAccepted } from "@/hooks/useIsUserRoleAccepted";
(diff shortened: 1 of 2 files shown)


src/components/WalletBalancesOverview.tsx
```diff
@@ -53,9 +54,21 @@ const WalletBalanceRow = ({
...
+          style={{
+            display: "inline-block",
+            width: "0.5rem",
+            height: "0.5rem",
+            borderRadius: "999px",
+            flexShrink: 0,
+            backgroundColor: accountColor(walletId),
+          }}
+          aria-hidden="true"
+        />
```
quietbits: We should never use inline styles. If currently there are any inline styles, that's an oversight and we should fix that.

There is a `Box` component for flex containers. We can use CSS variables to set the custom `backgroundColor` (`Box` component has an example of this).

Please let me know if you need help with anything. 🙏

hypekostas: @quietbits Thanks for the review!

Currently, there are 5 existing uses of the color dot that are inline, using no shared component. Therefore, inline style has been used here for consistency. An follow-up refactoring PR is intended to consolidate these into the stylesheet (out of scope for this PR).

As this PR involves a low-priority cosmetic change, it can be postponed until the cleanup refactor has taken place, then this branch can be updated accordingly. Actually, leading with the refactor makes more sense.

Several follow-up PR's are planned to clean up the frontend, including instances of inline styling that were added with the addition of the large multi-distribution-account feature.

quietbits: Thanks for adding more context. Honestly, I'm not a fan of keeping existing bad code for consistency. Ideally, it should be fixed as it is used because it's easy to forget about these things and they just keep piling up. Scout rule would be nice here: leave code better than you found it. :slightly_smiling_face:

quietbits (approved):

src/components/WalletBalancesOverview.tsx
hypekostas: I agree! PR #583 is up with the promised refactoring. Once that is merged, this branch will be updated accordingly.

hypekostas:
Rewritten on top of develop now that #583 landed: the branch is a single commit using the shared `DistributionAccountLabel` (dot + DEFAULT badge). The earlier inline dot, `.AccountBadge` class and switcher edit are all subsumed by the shared component.

PR #583, opened 2026-09-01
SDP-2138 refactor: consolidate distribution account labels into shared component

hypekostas:
### What

- Add `DistributionAccountLabel`: the single implementation of the account color dot + name, with an optional default marker (`"badge"` or `"text"`) and `(archived)` note. 
   - The color is the only dynamic value and is passed as a CSS custom property (`- DistributionAccountLabel-dot-color`), the same pattern `Box` uses.
- Replace the eight hand-rolled dot copies with it: account switcher (×3), Payments/Disbursements "Account" column (`SourceAccount`), distribution account heading, direct payment "Sending from" banner, disbursement wizard (×2). Delete their per-file dot/badge SCSS.
- Add `distributionAccountDisplayName()` for the `"Name (default)"` string used in `<option>` elements and string props.
- Rename `accountColor` to `distributionAccountColor`.
- One dot size (10px) everywhere; `develop` had 8px in the switcher/wizard and 10px elsewhere.

### Why

`develop` had nine independent implementations of the same dot with drifting sizes (0.5 vs 0.625rem), radii (`999px` vs `50%`) and spacing, and the `(default)` marker was rendered three different ways in nine places. Each new surfacewas copying one of them. One component removes the inline styles and makes the next surface a one-liner.

### Known limitations

- The palette in `distributionAccountColor.ts` is still eight hardcoded hex values; moving it to `--sds-clr-*` tokens (theme-aware, no inline style at all) is a separate PR that may need design input.
- `SettingsTeamMembers.tsx` carries a small unrelated cleanup so it can pass the pre-commit hook: `hideModal` moved above the effects that use it (pre-existing react-hooks/immutability error), the repo-standard set-state-in-effect disables on those calls, and import/order auto-fixes.

### Checklist

- [ ] Ready for production

### DIFF
--- a/src/components/DistributionAccountLabel/styles.scss
+++ b/src/components/DistributionAccountLabel/styles.scss
@@ -3,10 +3,16 @@
 .DistributionAccountLabel {
   display: inline-flex;
   align-items: center;
+  min-width: 0;
+
+  &__name {
+    min-width: 0;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
 
   &__dot {
     display: inline-block;
-    // Centers dot on the text when the parent is not a flex container (ignored in flex).
     vertical-align: middle;
     width: pxToRem(10px);
     height: pxToRem(10px);
--- a/src/components/DistributionAccountLabel/styles.scss
+++ b/src/components/DistributionAccountLabel/styles.scss
@@ -1,0 +1,35 @@
+@use "../../styles/styles-utils.scss" as *;
+
+.DistributionAccountLabel {
+  display: inline-flex;
+  align-items: center;
+
+  &__dot {
+    display: inline-block;
+    // Centers dot on the text when the parent is not a flex container (ignored in flex).
+    vertical-align: middle;
+    width: pxToRem(10px);
+    height: pxToRem(10px);
+    margin-right: var(--sds-gap-sm);
+    border-radius: pxToRem(999px);
+    flex-shrink: 0;
+    background-color: var(--DistributionAccountLabel-dot-color);
+  }
+
+  &__badge {
+    margin-left: var(--sds-gap-sm);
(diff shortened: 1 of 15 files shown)


src/components/DistributionAccountLabel/styles.scss
```diff
@@ -0,0 +1,35 @@
...
+
+  &__badge {
+    margin-left: var(--sds-gap-sm);
+    font-size: pxToRem(10px);
+    text-transform: uppercase;
+    letter-spacing: 0.03em;
+    font-weight: var(--sds-fw-bold);
+    padding: pxToRem(1px) pxToRem(6px);
+    border-radius: pxToRem(999px);
+    // gray-06/gray-12 clears WCAG AA in both themes (gray-04/gray-11 fell below in light).
```
quietbits: 👍 We can remove the comment about previous values.

quietbits (approved):
Thanks for cleaning these up, @hypekostas ! 🙌

src/components/DistributionAccountLabel/styles.scss
```diff
@@ -0,0 +1,40 @@
...
+
+.DistributionAccountLabel {
+  display: inline-flex;
+  align-items: center;
+  min-width: 0;
+
+  &__name {
+    min-width: 0;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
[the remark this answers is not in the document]
hypekostas: If `nowrap` were baked into the shared `__name`, all consumers would be forced to trade wrapping for the horizontal page overflow. Prefer to add the rule to the specific consumers than to make it global. This is a separate PR.

src/components/DistributionAccountLabel/styles.scss
hypekostas: Addressed!

